PR #3, opened 2026-09-04
fix: dangling examplenpu gitlink

vegah:
# Fix the submodule metadata: a fresh clone cannot build

Three files, ten lines. `git submodule update --init` — the first thing a new
contributor runs — **fails outright**, and the build cannot configure.

```
$ git clone https://github.com/Atomic-Germ/OpenFlowLM-Next
$ cd OpenFlowLM-Next
$ git submodule update --init --recursive
fatal: no submodule mapping found in .gitmodules for path 'docs/ExampleNPU'
```

---

### 1. `docs/ExampleNPU` is a gitlink with no `.gitmodules` entry

It is recorded in the index as mode `160000` (a submodule) but named nowhere in
`.gitmodules`. Git refuses to operate on **any** submodule while such an entry
is present, so the command above never gets as far as the ones that matter.

It is a leftover rather than something to restore: it was the tree the open
embedding engine was ported *from*, and that port has landed —
`src/open_embedding/{engine,npu_matmul}.cpp` are in the build. The prose in
`REPLACEMENT_PLAN.md` and `docs/precompiled_replacement_map.md` that cites paths
under it is historical and reads fine as history.

### 2. `.gitmodules` describes a state that does not exist

With the gitlink removed, `git submodule status` runs — and lists **nothing**,
because none of the four entries in `.gitmodules` is actually a submodule:

| entry | reality |
|---|---|
| `utilities/flm-add` | vendored — 6 tracked blobs in this repository |
| `utilities/q4nx-build` | vendored — 51 tracked blobs |
| `utilities/flm-test` | vendored — 15 tracked blobs |
| `third_party/tokenizers-cpp` | **does not exist at all** |

The last one is the build failure. `src/CMakeLists.txt:231`
`add_subdirectory()`s `../third_party/tokenizers-cpp` unconditionally and line
553 links `tokenizers_cpp` into `flm`, but nothing brings it in: `.gitmodules`
names it and no gitlink for it was ever committed, so
`git submodule update --init` fetched it and everything else silently.

**That is why the two are one PR.** The dangling gitlink made `git submodule` a
command nobody could run, which is exactly the command that would have reported
the missing one.

---

## What this does

* Removes the `docs/ExampleNPU` gitlink from the index.
* Adds `third_party/tokenizers-cpp` as a **real** submodule
  (`mlc-ai/tokenizers-cpp` @ `v0.1.1-11-gc586c52`, which pulls `msgpack` and
  `sentencepiece` in turn).
* Drops the three vendored directories from `.gitmodules`, so it describes what
  is there. Nothing is moved or deleted — they stay exactly as they are, as
  ordinary tracked files.

## Verified

```
$ git submodule update --init --recursive
Submodule 'msgpack' registered for path 'third_party/tokenizers-cpp/msgpack'
Submodule 'sentencepiece' registered for path 'third_party/tokenizers-cpp/sentencepiece'
...
$ git submodule status
 c586c52f93f7b060753bd2388eb96a105cb7374d third_party/tokenizers-cpp (v0.1.1-11-gc586c52)

$ cmake -S src -B src/build ...
-- Configuring done
-- Generating done
$ cmake --build src/build --target flm
[317/317] Linking CXX executable flm.exe
```

`flm help` is byte-identical before and after: this changes no code.

## One environment note, not a repo problem

On Windows without Developer Mode, `sentencepiece`'s CMake fails at

```
file failed to create symbolic link '.../sentencepiece/third_party/absl':
A required privilege is not held by the client.
```

Its own `if (NOT EXISTS ...)` guard means a **directory junction** placed there
first satisfies it, and `mklink /J` needs no privileges:

```
mklink /J <repo>\third_party\tokenizers-cpp\sentencepiece\third_party\absl ^
          <repo>\third_party\tokenizers-cpp\sentencepiece\third_party\abseil-cpp\absl
```

That is upstream sentencepiece's behaviour, not this repository's, and is noted
here only so the next person does not lose an hour to it.

---

*Found while building this fork to add an embedding backend; that work is a
separate PR, which depends on this one because without it the tree does not
configure.*

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,3 @@
 [submodule "third_party/tokenizers-cpp"]
 	path = third_party/tokenizers-cpp
 	url = https://github.com/mlc-ai/tokenizers-cpp
-[submodule "utilities/flm-add"]
-	path = utilities/flm-add
-	url = https://github.com/Atomic-Germ/flm-add
-[submodule "utilities/q4nx-build"]
-	path = utilities/q4nx-build
-	url = https://github.com/Atomic-Germ/q4nx-build
-[submodule "utilities/flm-test"]
-	path = utilities/flm-test
-	url = https://github.com/Atomic-Germ/flm-test


Atomic-Germ:
That's weird considering `git clone git@github.com:Atomic-Germ/OpenFlowLM-Next --recursive` "worked for me", I was entirely unaware of this. I assume I was compensating with environment variables and the like.

PR #6, opened 2026-09-05
feat: Qwen3.6-MoE decode on open XDNA2 kernels

### LINKED ISSUE #5: Open XDNA2 kernels for Qwen3.6-MoE decode
Continuing the open-kernels thread from ROCm/FastFlowLM#690 — @Atomic-Germ
asked (2026-09-04, in that issue) for this to land as an issue referenced from
a PR, so here it is, with a PR attached rather than a proposal: **Qwen3.6-MoE
(35B-A3B: hybrid gated-DeltaNet / full-attention / 256-expert MoE) decoding
end to end on hand-written IRON/mlir-aie kernels, with the closed
`qwen3_6_moe_npu` engine no longer in the loop.**

PR: (opening alongside this issue)

## What's in it

- **21 kernel designs** (`open_kernels/designs/`) covering every op in a
  decode step — q4_1 GEMV with dequant fused into the integer matrix unit,
  q8 lm_head, gated DeltaNet, the conv1d/SiLU/qk-norm glue, full attention
  with a dynamically-patched KV window, the MoE router with on-device top-8 —
  plus fused whole-layer designs (`layer_x/`) that run either layer type as
  two dispatches on one xclbin context.
- **A host harness** (`open_kernels/harness/`) that drives them from a 6-line
  `.cfg` language over the same XRT route `npu_matmul.cpp` already uses, plus
  three instruction-patching directives (`moeroute2`, `attnpos`) that let one
  compiled program serve every layer and every token — no per-step
  recompilation, ~0.04ms per patch.
- **An open weight-container reader + packer** (`open_kernels/model/`,
  `src/open_qwen36/`) for FastFlowLM's `.q4nx` format, replacing
  `q4_npu_eXpress.dll` on this path, plus an independent fp64 CPU
  implementation of the whole decode step used as the correctness oracle.
- **A resident engine** (`src/open_qwen36/`) implementing the app's own
  `causal_lm` interface, so the existing tokenizer/sampler/chat-template/
  server code drives it exactly as it drives the closed engine.
  `Qwen3_6_MOE::load_model` picks it automatically whenever the kernels are
  installed for the model; `FLM_QWEN36_ENGINE=open|closed` forces either way.

## Verified

- All 40 layers of stock `Qwen3.6-35B-A3B-NPU2` against the fp64 oracle:
  logits corr 0.999997, same argmax and top-5, every layer's residual
  ≥ 0.99992. The on-device router agreed with the oracle's own top-8 in 38 of
  40 layers; the two exceptions are 8th-of-8 ties with an expert-probability
  gap of ~8e-5, below bf16 resolution at that magnitude.
- Through the real app, built from scratch on a machine with none of the
  toolchain installed (`src/build-windows-vcpkg.cmd`): `flm serve
  qwen3.6-moe:35b-a3b` selects the open engine and answers chat completions
  correctly and deterministically at 5–7 tok/s decode, 1.4–2.2s TTFT — in the
  same range as the closed engine on the same hardware.

Full numbers, per-kernel timing and the traps hit along the way are in the
PR and in `src/open_qwen36/README.md` / `open_kernels/README.md` once it
lands.

## Two things worth deciding here rather than in review

1. **Toolchain: this pins mlir-aie 1.4.2.** 1.3.4, as this repo currently
   configures it, can't even `import aie.iron` (a namespace collision with
   `eudsl-python-extras`), and its IRON runtime API is different enough
   (`rt.task_group()` vs free-standing `TaskGroup()`, `rt.fill(fifo, ...)` vs
   `fifo.fill(...)`) that none of these designs run on it. The one existing
   design on the old API (`npu_offload/matmul/matmul_whole_array.py`) is
   ported in the first commit; all 12 embedding matmul shapes still rebuild
   bit-identical to the shipped binaries. If there's a reason to stay on
   1.3.4 I'd want to hear it before more designs pile onto 1.4.2-only APIs.
2. **Weight format.** This still reads FastFlowLM's `.q4nx` container
   (@Atomic-Germ noted in #690 that q4nx isn't the intended long-term
   format). The packer's transform is a byte permutation plus fp16→bf16
   narrowing — nothing is dequantized or requantized — which is what would
   let it read GGUF Q4_1/Q8_0 directly instead; `open_kernels/q4_1_pack.py`
   is a first cut at that direction, unused by the engine yet. Worth agreeing
   on GGUF as the target before it's load-bearing in more places.

## Not yet done

- **Batched prefill.** Prompt tokens go through the decode step one at a
  time (exact for this architecture, since each layer's state update only
  ever sees one token) at ~124ms/token on the full model. The closed engine
  has real batch-prefill kernels; nobody has written open ones.
- **Vision.** The model is a VLM; the open engine has no encoder and defers
  to the closed one for image inputs.
- Attention cost at long context is untuned (~24µs per cached row per
  attention layer) — not a hard limit, just not looked at yet.

## Credit

The kernel package is vendored from my own prior work (`phlegm`,
`tools/open-kernels/`), full provenance and commit pin in
`open_kernels/PROVENANCE.md`. The GEMV tile arithmetic and the `rot13` smoke
test build on @vegah's
[LLMNpuTest](https://github.com/vegah/LLMNpuTest) (Apache-2.0, notice
preserved in the relevant file headers) — same lineage as the embedding
kernels already in this repo.

Cyronius:
# Qwen3.6-MoE decode on open XDNA2 kernels

Refs #5

Replaces the closed `qwen3_6_moe_npu` engine with hand-written IRON/mlir-aie
kernels for Qwen3.6-35B-A3B-NPU2's hybrid architecture (gated DeltaNet linear
attention, full attention, 256-expert MoE), wired all the way into the app as
a real `causal_lm` implementation — not a batch test harness. Commits, each
independently buildable/verifiable:

1. `ironvenv: pin mlir-aie 1.4.2, port IRON designs off the 1.3.4 runtime API`
2. `open_kernels: open XDNA2 kernels for a Qwen3.6-MoE decode step`
3. `add open kernels for Qwen3.6-35B-A3B-NPU2` — the compiled binaries; **removed again in 5**
4. `open_qwen36: resident Qwen3.6-MoE engine on the open kernels`
5. `open_kernels: ship the Qwen3.6 kernels as source, portable fixture paths` — review follow-ups (the conventions from #4, Copilot's review), see §3 and §6

## 1. Toolchain pin

1.3.4, as this repo currently configures it, can't `import aie.iron` at all —
the `eudsl-python-extras` wheel installs a second top-level `aie/` package
that never merges with `mlir_aie`'s real one (`ModuleNotFoundError: No module
named 'aie.extras.util'`). With that fixed by hand, 1.3.4 still can't run
these designs: the IRON runtime API changed shape between the releases
(`rt.task_group()` → free-standing `TaskGroup()`, `rt.fill(fifo, ...)` →
`fifo.fill(...)`). `ironvenv-requirements.txt` now pins `mlir_aie==1.4.2` +
matching Peano and drops the eudsl line; the one design in this tree on the
old idiom (`matmul_whole_array.py`) is ported.

**Verified:** all 12 embedding matmul shapes rebuild under 1.4.2 (~20s each,
25% smaller — 69KB vs 93KB). Two rebuilt shapes are bit-identical to the
shipped xclbins (maxabs 6.6e-7 vs the fp64 reference), so the shipped
binaries are left in place — the port is proven without churning 24 binary
files.

## 2. The kernels

21 designs (`open_kernels/designs/`) covering every op in a decode step —
q4_1 GEMV with dequant fused into the integer matrix unit, q8 lm_head, gated
DeltaNet, the conv1d/SiLU/qk-norm glue, full attention with a
dynamically-patched KV window, MoE routing with on-device top-8 — plus fused
whole-layer designs (`layer_x/`) that run either layer type as two
dispatches on one xclbin context. Provenance (vendored from my own prior
work, `phlegm`), license notices (MIT + vegah's Apache-2.0 notice on the GEMV
tile arithmetic, same lineage as this repo's existing embedding kernels) and
what was deliberately left out (compiled binaries, captured fixtures, run
configs — all gitignored) are in `open_kernels/PROVENANCE.md`.

New in this tree:

- **`open_kernels/harness/run_kernel.cpp`** — an N-buffer opcode-3 runner
  (`xrt::xclbin → hw_context → xrt::kernel`, the route `npu_matmul.cpp`
  already uses) that speaks a 6-line `.cfg` language, so every design's
  `make_test.py`/`compare.py` pair runs against it unchanged.
  **`harness/stream_patch.hpp`** factors out three directives that let one
  compiled instruction stream serve every layer and every token instead of
  being rebuilt per step:
  - `moeroute`/`moeroute2` re-point a kernel's expert weight fills at
    whatever the on-device router just chose
  - `attnpos` rewrites the KV window length, new-row offset and RoPE record
    offset for the current position

  All three patch one word of an mlir-aie instruction stream (op `0x81`, a
  DDR-patch: register, buffer arg index, byte offset) plus one
  instruction-BO sync — **~0.04ms per patch**.
- **`open_kernels/model/`** — an open reader for FastFlowLM's `.q4nx`
  container (`q4nx.py`, 1.0.2 q4_1 format only), the NPU pool/pack/side
  packer (`pools.py`, checked byte-for-byte against a captured-buffer
  reference for both layer types), an independent **fp64 CPU implementation
  of the whole decode step** (`replica.py`) as the correctness oracle, and
  `make_decode.py`/`compare_decode.py` to build a driver program for an
  arbitrary layer count and score it against that oracle.
- **`designs/gemv_q4/make_test.py`, `designs/ln/make_test.py`** — synthetic
  Q4_1 fixtures by default (random blocks, an fp64 reference derived from
  the same bytes), so correctness can be checked with no captured model and
  no NPU-resident weights.
- **`q4_1_pack.py`** — GGUF-style Q4_1 blocks to pool-order chunks, with a
  self-test. The transform is a byte permutation plus fp16→bf16 narrowing —
  nothing is dequantized or requantized — which is what would let this
  become the packer for GGUF-sourced weights directly (see the format
  question in #5).
- **`fixture_paths.py`** (commit 5) — the one place the generators get
  paths from. Every generated `run.cfg` is relative to its design directory
  (`run_kernel` resolves against the cfg's directory); fixtures that slice
  weights out of captured FLM buffers read `OPEN_KERNELS_CAPS` and stop with
  a message when it is unset; the `*_chain` harnesses (phlegm's step-by-step
  chains, kept for the record) take `PHLEGM_KERNEL_INTERP` / `MODEL_Q4NX`.
  No path in the tree points at a developer's checkout any more.

**Verified** on stock Qwen3.6-35B-A3B-NPU2: the full 40-layer decode step
matches the fp64 oracle at **logits corr 0.999997, same argmax and top-5,
every layer's residual ≥ 0.99992**. The on-device router agreed with the
oracle's own top-8 in 38 of 40 layers; the two exceptions are 8th-of-8 ties
with an expert-probability gap of ~8e-5, below bf16 resolution at that
magnitude. Full per-kernel timing is in `open_kernels/model/README.md`.

## 3. The compiled kernels — built, not checked in

Following the convention settled on #4 (xclbins ship as source; pre-built
only in the distributed package), the six kernel sets the engine loads are
**not** in the repository: `src/xclbins/Qwen3.6-35B-A3B-NPU2/open_kernels/`
is gitignored with the same note as the BERT design sets. What is in the
repository is the source (`open_kernels/designs/{layer_x,ln,lm_head_q8}`) and
the one command that produces them:

```
source ~/ironenv142/bin/activate      # mlir-aie 1.4.2 + Peano, xclbinutil/aiebu-asm on PATH
python open_kernels/export_qwen36_kernels.py
#  -> src/xclbins/Qwen3.6-35B-A3B-NPU2/open_kernels/{lx0,lx1,ax0,ax1,ln,lm_head_q8}/ + toolchain.json
```

~6 minutes (four ~90 s `layer_x` builds). `toolchain.json` records the
mlir-aie/Peano versions, the tree's commit and every sha256. Knobs per set,
lookup locations and the install path are in `src/open_qwen36/README.md`;
`.opencode/skill/open-qwen36-kernels/SKILL.md` is the skill AGENTS.md asks
for per xclbin set.

**Is the source really the source?** `export_qwen36_kernels.py --check DIR`
compares a fresh build against a previous one. Run here against the binaries
commit 3 had shipped (built 2026-09-04, same toolchain):

| | result |
|---|---|
| 6 × `insts.bin` | **byte-identical** |
| 6 × `final.xclbin` | identical apart from 78–82 bytes each: the axlf header's unique id / timestamp / UUID, the PDI UUID in `AIE_PARTITION`, the boot-image header's unique id and its checksum, and xclbinutil's mirror-JSON tail |

`--check` masks exactly those stamp fields by parsing the axlf and partition
structs and fails on any other byte (verified with a flipped CDO byte). A
rebuilt `ln` was also run on the NPU through the harness: same result as the
shipped one (`y` maxrel 5.4e-8, `xn` bit-exact).

## 4. The engine

`src/open_qwen36/` wires the kernels into a resident `causal_lm`
implementation:

- **`q4nx_file.{hpp,cpp}`** — an open, mmap'd `.q4nx` reader (refuses 1.0.3/
  Q4_K rather than misreading it), replacing `q4_npu_eXpress.dll` on this
  path.
- **`pools.{hpp,cpp}`** — the same packing as `open_kernels/model/pools.py`,
  ported to write straight into a resident device buffer.
- **`core.{hpp,cpp}`** — `Core`: opens the device and the six kernel
  contexts once, packs every layer's weights directly into resident BOs
  (**84s for the full 40-layer model**), then `step(token, want_logits)`
  runs one decode step. Prompt tokens are prefilled by sequential decode
  (exact for this architecture — each layer's state update only ever sees
  one token at a time). `checkpoint()`/`restore()` copy the linear-attention
  state and the KV rows written so far off/onto the device, for the app's
  prompt-cache reuse.
- **`engine.{hpp,cpp}`** — `Engine : causal_lm`: `forward`/`prefill`/
  `set_context_length`/`load_weights`/`clear_context`/`get_k_cache`/
  `get_v_cache`/`get_current_context_length`/`checkpoint`/`restore`.
  `find_kernels()` looks for `xclbins/<model>/open_kernels/` (installed),
  `<model dir>/open_kernels/` (a model shipping its own), or
  `FLM_OPEN_KERNELS_DIR`. An image payload is refused with a clear error —
  there is no open vision path yet.
- **`cli.cpp`, `chat.py`** — drive `Core` without the app (ids in, tokens/
  logits out). This is the actual test surface for everything below: the app
  itself doesn't build in every environment, and `cli.cpp` links the
  identical `core.cpp`/`pools.cpp`/`q4nx_file.cpp` the app does.
- **`modeling_qwen3_6_moe.{hpp,cpp}`, `CMakeLists.txt`** —
  `Qwen3_6_MOE::load_model` uses the open engine whenever its kernels are
  found for the model, the closed DLL otherwise;
  `FLM_QWEN36_ENGINE=open|closed` forces either way. `checkpoint()`/
  `restore()` now go through the `causal_lm` base instead of a
  `qwen3_6_moe_npu`-specific call, since both engines implement it. XRT
  builds only (`FLM_USE_HRX=OFF`), same restriction as `open_embedding`.

### Verified end to end

| test | result |
|---|---|
| `cli.cpp`, 8/40 layers, 3 greedy tokens | logits corr 0.999998/0.999996/0.999991 vs the fp64 oracle — identical to the batch harness |
| `cli.cpp`, all 40 layers, chat prompt | coherent text, ~7 tok/s decode, 124ms/token prefill |
| `cli.cpp`, repeated request + checkpoint/restore | byte-identical output, 6 times in a row |
| **Through the real app** (`src/build-windows-vcpkg.cmd`, built from scratch — see `WinSetup.md`) | `flm serve qwen3.6-moe:35b-a3b` selects the open engine; **8/8** OpenAI-style chat completions, correct and deterministic (temperature 0, repeated prompts match), **5.3–6.9 tok/s decode, 1.4–2.2s TTFT** |

### One real finding, and its fix

The first server session hung a kernel (ERT state 8) after four requests and
every submission failed afterward. The box was down to 0.6GB free with
`flm`'s working set at 35GB — the engine's 21.6GB of resident NPU buffers
plus ~13GB of the `.q4nx` mapping's pages the packer had touched and never
released. Windows manages NPU buffers through the video memory manager and
can evict them under memory pressure; a submission that can't page its pool
back in hangs, and the hardware context doesn't recover on its own. Two
fixes are in this PR:

- `Q4nxFile::drop_pages()` unmaps and remaps the container right after
  packing (**35GB → 21.5GB** working set; 16.6GB stayed free through a clean
  8-request run afterward)
- `Engine::ensure_alive()` rebuilds a poisoned engine on the next request
  (~90s) instead of leaving it permanently broken

Budget ~25GB free for the full model.

## 5. Not yet done

- **Batched prefill.** ~124ms/prompt token, one token at a time. The closed
  engine has real batch-prefill kernels; nobody has written open ones yet
  (this is weeks of work gated on one experiment, not a quick follow-up).
- **Attention cost at long context** (~24µs/cached row/attention layer) is
  untuned — not a hard limit (the KV/ptab buffers are just sized off the
  app's context length now, no more fixed cap), just not looked at yet.
- **Vision.** The model is a VLM; the open engine has no encoder and defers
  to the closed one for image inputs.

## 6. Review follow-ups (commit 5)

From the conventions on #4:

- **xclbins ship as source** → §3. The binaries are gone from the tree, the
  directory is gitignored, the export script and the reproducibility check
  replace them.
- **Tests belong in flm-test** → this PR adds no standalone endpoint test
  scripts. The per-design `make_test.py`/`compare.py` pairs are kernel unit
  tests against fp64 references; `chat.py` drives the standalone CLI. The
  app path is exercised with `flm-test --llm` like any other model — run
  here against `flm serve qwen3.6-moe:35b-a3b` on the open engine
  (2026-09-05, `utilities/flm-test` at this tree): **PASS**, both streamed
  turns coherent and on topic (Maxwell's equations, then why they matter).
  The numbers it logged are the honest state of §5: turn 1, 15 prompt
  tokens → TTFT 1.8 s, 571 tokens at 4.4 tok/s; turn 2 carries the 587-token
  history → TTFT 123 s (sequential prefill, 4.8 tok/s) and 2.8 tok/s decode
  at ~1000 KV rows (the untuned per-row attention cost).

From Copilot's review:

- `build_mlir.sh`: `aiecc … | tail` could hide a failed build → `set -euo pipefail`.
- `router/make_test.py`, `silu_mul/make_test.py` (and the suppressed
  `router:40`): hard-coded `/mnt/c/caps` and `C:/code/phlegm` paths → fixed
  in **every** generator that shared the pattern (17 scripts), via
  `fixture_paths.py`: relative cfg paths, `OPEN_KERNELS_CAPS` with a clear
  error, `PHLEGM_KERNEL_INTERP`/`MODEL_Q4NX` for the chain harnesses.
- Same class, not flagged: `harness/build.cmd` and `open_qwen36/build.cmd`
  no longer default the XRT header/lib dirs to a phlegm checkout (they stop
  and name the two variables); `model/make_decode.py` and `chat.py` default
  the model dir to `~/.flm/models/<model>` rather than one user's home.

## Build notes

`src/build-windows-vcpkg.cmd` is the path that got a completely bare Windows
box (no Boost, no vcpkg, no `third_party/` populated) building `flm.exe`:
vcpkg classic mode for Boost/curl/FFmpeg/FFTW3, a Visual Studio instance with
the C++ toolset (Community without it is found first by the generator and
fails), and a directory junction for `sentencepiece`'s symlink (which needs a
privilege a normal console session lacks). Traps and the one-time setup are
in `WinSetup.md`. Also worth knowing before running it: the app resolves its
home to the *system* profile from a non-interactive shell
(`FLM_MODEL_PATH=%USERPROFILE%\.flm` fixes it), and `flm serve`/`run` delete
and re-download a model that fails its registry `flm_min_version` check —
relevant if testing against a `.q4nx` built for an older FLM version.

---


Cyronius:
Pushed `d96a918` — review follow-ups, applying the conventions from #4 and Copilot's review. The PR description is updated (§3, §6).

**Convention from #4: xclbins ship as source, pre-built only in the distributed package.** Commit 3 had checked in the six kernel sets under `src/xclbins/Qwen3.6-35B-A3B-NPU2/open_kernels/`. They are removed and the directory is gitignored (same note as `src/xclbins/BERT-h*/`). `open_kernels/export_qwen36_kernels.py` builds them from `designs/` in one command (~6 min, WSL, mlir-aie 1.4.2) and writes a `toolchain.json`; `--check DIR` compares a rebuild with a previous one. Checked against the removed binaries: all 6 instruction streams byte-identical, all 6 xclbins identical apart from 78–82 bytes of per-build stamps (axlf unique id / timestamp / UUID, PDI UUID, boot-image header unique id + checksum, the mirror-JSON tail — masked by parsing the structs, and the check fails on any other byte). A rebuilt `ln` also ran on the NPU with the same result as the shipped one. Documented in `src/open_qwen36/README.md` and a skill file per AGENTS.md.

**Convention from #4: tests live in flm-test.** No standalone endpoint test scripts in this PR; the `make_test.py`/`compare.py` pairs are kernel unit tests against fp64 references, and the app path is exercised with `flm-test --llm` like any other model. Ran it here against `flm serve qwen3.6-moe:35b-a3b` on the open engine: **PASS** (both streamed turns coherent). Its timings are the state of the "not yet done" list: 15-token prompt → TTFT 1.8 s, 4.4 tok/s over 571 tokens; the 587-token second turn → TTFT 123 s (sequential prefill) and 2.8 tok/s at ~1000 KV rows (untuned per-row attention cost).

**Copilot:**
- `build_mlir.sh` — `set -euo pipefail`, so a failed `aiecc` behind the `| tail` is no longer reported as `BUILD_MLIR_OK`.
- `router/make_test.py`, `silu_mul/make_test.py`, suppressed `router:40` — fixed, and the same fix applied to every generator that shared the pattern (17 scripts), through one helper `open_kernels/fixture_paths.py`: generated `run.cfg` paths are relative to the design directory (`run_kernel` resolves against the cfg's directory); captured FLM buffers come from `OPEN_KERNELS_CAPS` (plus `OPEN_KERNELS_CAPS_HOST` for a WSL-generator / Windows-driver split) with a clear message when unset; the `*_chain` harnesses take `PHLEGM_KERNEL_INTERP` / `MODEL_Q4NX`.
- Same class, not flagged: `harness/build.cmd` and `open_qwen36/build.cmd` no longer default the XRT header/lib dirs to a phlegm checkout; `model/make_decode.py` and `chat.py` default the model dir to `~/.flm/models/<model>`.

vegah (approved):

vegah:
I would just merge it and then start the work of fixing stuff, one by one if something doesn't work :) Right now we have a lot of huge commits, and this makes it a little chaotic :)

Atomic-Germ:
Y'all are fast

PR #10, opened 2026-09-05
build: one way to build the xclbins, and one contract between the pro…

vegah:
…ducers

Two producers build design sets in this repository -- npu_offload/gemm_rtp/ (five BERT/embedding GEMM families) and open_kernels/ (six Qwen3.6-MoE decode sets). They arrived through different pull requests and each invented the same conventions separately: build from source, do not check the binaries in, drop a toolchain.json beside them, offer a --check that proves the source really is the source. Independently invented conventions drift, and three had already drifted:

  * toolchain.json named the same file with DIFFERENT schemas -- three fields from gemm_rtp, eight from open_kernels, both landing under src/xclbins/. src/open_npue/npu_device.cpp reads it and reports "unavailable" for anything it does not find, so the disagreement degraded silently instead of failing.
  * The IRON build cache (~/.npu/cache) is GLOBAL and gemm_rtp's purge() deletes entries by CONTENT markers, so two builds delete each other's work -- guaranteed, not unlucky: qkv and attn_out depend on neither --gated-ffn nor --intermediate, so the two hidden-768 families share 8 of their 16 entries exactly. build.ps1's own header claimed "export_gemm_rtp.py holds a lock now and refuses in under a second". In this tree it does not: that copy predates the upstream change. Meanwhile open_kernels became a second client of the same cache.
  * mlir-aie was looked for in three places by three files (C:\dev\mlir-aie, ~/ironenv142, ironvenv + utilities/mlir-aie), so a from-scratch build followed whichever document its reader found first.

Nothing here merges the two producers -- they build genuinely different things -- and NOT ONE FLAG MOVES. families.json and export_qwen36_kernels.py's SETS remain the only places their own knobs are written down; the new code reads both and restates neither.

New:

  tools/npu_designs.py    everything two producers must agree about: the
                          toolchain.json schema, the shared cache lock, the npu2
                          device pin, the xclbin comparison (moved out of
                          export_qwen36_kernels.py, because "did this source
                          produce these bytes?" is the only honest test of a
                          repo that ships source, and both producers make that
                          claim), and a registry that reads each producer's own
                          spec.
  tools/build_designs.py  doctor | list | build | check -- one command, both
                          platforms, one lock held across both producers.
  docs/design-sets.md     the output convention, the schema, and the one gap
                          this does not close, stated rather than hidden.

Changed:

  toolchain_provenance.py     now a shim over the shared writer. The three
                              fields it wrote keep their names, so npu_device.cpp
                              is unaffected; five more are added.
  export_qwen36_kernels.py    takes the shared lock (--force-unlock), uses the
                              shared hash/compare/provenance; 73 lines of
                              duplicated code deleted. --no-build deliberately
                              takes NO lock: it is what you run while a build is
                              going.
  build_design.py             shared device pin.
  build.ps1                   a wrapper over the entry command, and its false
                              claim about the lock corrected. It also stops with
                              the dot-source instructions when there is no
                              `python` at all, rather than "the term 'python' is
                              not recognized".

Verified by building all five BERT families from an empty src/xclbins through the new entry command on this machine (Windows, C:\dev\mlir-aie IRON env): five families, 20 files each, `check` green against families.json, and every set carrying the new toolchain.json. The lock was visibly held by build_designs.py throughout.

doctor earned its place on the first run: it reports mlir_aie pinned 1.4.2, installed 1.4.2.dev16+g7e00b57. The sets on this machine are built by a dev build, not the pinned release. The old toolchain.json recorded the version and nothing ever compared it to the pin.

Not verified here: the six Qwen sets. They need WSL with xclbinutil/aiebu-asm and were not rebuilt for this change; their code paths are exercised only by import, --help and the shared-module unit checks.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,23 @@ Note: Peano (llvm-aie) has not been added to PATH to avoid conflict with
 
 Activate the ironvenv/bin/activate; use source utilities/mlir-aie/utils/env_setup.sh also *if needed*
 
+## Building xclbins
+
+The xclbins are built, not checked in. There is ONE command for every design set
+in the repository, and one document describing the convention:
+
+    python tools/build_designs.py doctor   # can this shell build at all?
+    python tools/build_designs.py list     # what exists, and is it built?
+    python tools/build_designs.py build    # build what is missing
+    python tools/build_designs.py check    # do the built sets match their spec?
+
+Read `docs/design-sets.md` before adding a third producer. The short version:
+`ironvenv-requirements.txt` is the only toolchain pin, `tools/npu_designs.py`
+holds everything two producers must agree about (the `toolchain.json` schema,
+the shared `~/.npu/cache` lock, the device pin, the xclbin comparison), and each
+producer's flags stay in its own spec — `npu_offload/gemm_rtp/families.json` and
+`open_kernels/export_qwen36_kernels.py`'s `SETS`. Nothing restates them.
+
 <available_skills>
   <skill>
     <name>npu_offload_pipeline</name>
--- a/build_executable.bat
+++ b/build_executable.bat
@@ -1,0 +1,49 @@
+@echo off
+setlocal
+::
+:: Build flm.exe on Windows.
+::
+::     build_executable.bat              configure + build via vcpkg + MSVC
+::     build_executable.bat <vcpkg-root> use a vcpkg other than C:\dev\vcpkg
+::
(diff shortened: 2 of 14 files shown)


vegah:
Draft pr for starting unifying the build process between embedding etc.

Atomic-Germ:
I've got another way; I'm going to make a separate PR and then we can all discuss which way to go forward. Give me a bit.

Issue #18, opened 2026-09-06
Support Hy-MT2-7B (HunYuan dense) on the open kernels


Cyronius:
Someone asked whether Hy-MT2-7B can run on the open XDNA2 kernels. It can, and it is close to free, because the model is shaped exactly like the Llama 3.1 8B point we already run.

## What the model is

[tencent/Hy-MT2-7B](https://huggingface.co/tencent/Hy-MT2-7B) is Tencent's HunYuan V1 dense translation model — 33 languages, plus a handful of Chinese dialect and minority-language pairs. Architecture `hunyuan_v1_dense` (`HunYuanDenseV1ForCausalLM`), GGUF architecture `hunyuan-dense`.

Its `config.json` lines up with Llama 3.1 8B almost field for field:

| | Hy-MT2-7B | Llama 3.1 8B |
|---|---|---|
| hidden / layers | 4096 / 32 | 4096 / 32 |
| query heads / kv heads / head dim | 32 / 8 / 128 | 32 / 8 / 128 |
| FFN width | 14336 | 14336 |
| activation / norm | SiLU / RMSNorm, eps 1e-5 | SiLU / RMSNorm, eps 1e-5 |
| vocabulary | 128167, tied head | 128256, its own head |

So the dense recipe needs no new GEMV width, no new norm width and no new band split. Every kernel point it asks for is already in the catalogue's validated set, and `per_call` lands on 1 for the same reason Llama's does — the 14336-wide activation table is 32 KB.

## What is actually different

Three things, and none of them is a shape.

**The q/k RMSNorm weight is applied after the rotary embedding, not before.** HunYuan computes `query_layernorm(apply_rotary_pos_emb(q))`; Qwen3 and our `attn.h` normalize first. This matters because RoPE rotates the pairs `(i, i + d/2)` and those two positions generally carry different norm weights, so the multiply does not commute with the rotation. It does not change the scale, though — RoPE is orthogonal and the rotary dimension here is the whole head, so the RMS comes out the same either way. In other words, one multiply has to move to the other side of the rotation. That is the only kernel change, and it needs a hardware comparison before the new point can enter the catalogue.

**The RoPE scaling looks dynamic and is not.** The config asks for `{"type": "dynamic", "alpha": 1000}`, which both transformers and llama.cpp resolve once at load into a single stretched base, `rope_theta * alpha^(d/(d-2))` = 11158839.925. Nothing reaches the kernel but a different position table.

[...]

PR #20, opened 2026-09-06
linux build script - includes helper for iron toolchain, builds from new repo clone on linux

Atomic-Germ:
Just the single script, didn't touch anything else. It activates a 3.12 venv named ironvenv, activates necessary things for xrt, and has a switch for building all of the kernels (as of the moment I wrote it anyway)


Atomic-Germ:
Shouldn't be considered until #18 lands, then this gets updated

Cyronius (approved):
it looks good. This won't work well when we add a million more kernels, but gotta start somewhere. I have no idea what we're going to do for a build process when that process rebuilds kernels for 50 models. 

I'm sure you still have updates to do to this to add the granite and hy-mt2-7b and probably some merge conflicts, but it's approved.

Atomic-Germ:
I've got further work on this wrapped up in the gguf PR; I'll separate it out though.

PR #21, opened 2026-09-06
feat(open-kernels): IBM Granite 4.2 3B as the sixth(?) family on the dense recipe

vegah:
Granite 4.2 3B runs on the open XDNA2 kernels — no new engine, no new design,
  no new AIE code. `dense.py` already had everything; this is the spec
  derivation, the catalogue points and the registration.

  Granite is the first **head_dim 64** point in the tree, and the first at
  40 heads. `designs/attn/attn.h` already carried HD 64's `kScale`, and every
  `static_assert` passes at (NH 40, KVH 8, HD 64, ROT 64).

  ## Why it is legal

  Granite is Llama plus four scalar multipliers that `ModelSpec` cannot express
  and `attn.h` cannot apply — the same wall Gemma 3 hits with
  `query_pre_attn_scalar`. They do not need expressing, because q4nx-build folds
  all four into the weights: for 4.2-3B the only non-unit factor is
  `attention_multiplier = 0.015625` at hd 64, so `q_proj *= 0.125` and the folded
  config reads exactly `64**-0.5`. A power of two, so the fold is exact in bf16.

  An unfolded container is refused by name twice — at generation
  (`_granite_scale_check`, from HF config and GGUF alike) and at load, because
  `hf_config_check` carries `attention_multiplier`. An unfolded model does not
  crash; it returns plausible output.

  ## What it adds

  - `recipes/`: `_granite_hf` / `_granite_gguf`, the family in `families.py` and
    `dense.py`, `specs/granite42-3b.json` (derived from the published container,
    not transcribed)
  - `catalogue.py`: `attn.head_dim 64`, `attn.num_heads 40`, `gemv_q4.K 8192`
  - `specs/open-engine/`: `OPEN-FAMILY-GRANITE`, `tests/test_granite.py`
  - `AutoModel`: `modeling_granite.{hpp,cpp}`, the four `all_models.hpp` edits,
    `model_list.json` / `model_info.json`
  - the per-family tables in `make_decode.py` and `chat.py`
  - `.opencode/skill/open-granite-kernels/`

  ## Validated

  4-layer slice against the fp64 oracle: logits corr 0.999998 / 0.999990, same
  argmax, identical top-5, residual corr 0.999990–0.999999 every layer.
  `flm run granite:3b` selects the open engine and answers coherently.
  `specs/open-engine/tests`: 53 passed.

  ## New: OPEN-ATTN-CONTEXT

  Measuring Granite produced the first per-position breakdown anyone has run on
  this engine, and it is not a Granite property:

  | position | Qwen3-4B `part0` (36L) | Granite `part0` (40L) |
  |---:|---:|---:|
  | 0 | 56.7 ms | 60.0 ms |
  | 2048 | 3120.6 ms | 4121.6 ms |

  Linear in both — 41.6 and 49.6 µs per layer per position — while `lm_head`
  stays flat. At position 2048 one token costs 3–4 s. The cost is neither
  bandwidth nor arithmetic: Granite reads half the KV bytes per position per
  layer and does fewer MACs, yet its slope is 19% steeper. Head count is what
  tracks (40/32 = 1.25 against a measured 1.19), pointing at a fixed per-head
  cost in `attn_row_impl` that does not shrink when `head_dim` halves.

  Filed as an observation, with `open_kernels/model/sweep_positions.ps1` to
  reproduce it.

  ## Draft

  Landing the family is the ask here. The attention finding above is the
  follow-up, and it is worth more than this PR: it applies to all five families.


Cyronius:
Oh good. I have a bunch of model and model families in flight right now, but I believe this one and your other findings were work i had deferred so i could get all these qwen models working well. 

This is also one of the more practical models for the npu as opposed to the massive and slow beasts I've been on.

Atomic-Germ:
Nice; I love reusable code.

vegah:
The copilot review is hanging. Can someone review? :)

Cyronius:
other than the rebasing against main, this was uncovered:

2. utilities/q4nx-build has no Granite builder, and the whole PR depends on one. The spec, the two refusal paths and the tests all say the container must be folded "by q4nx-build" — but there is no q4nx/models/granite.py and no configs/granite.json in this PR or in main. Every one of the previous five families shipped both. As it stands, the only way to obtain a container this code accepts is to download the author's personal HF upload. Per the maintainer's own standard in CLAUDE.md ("adapt the existing tools; q4nx-build must stay in sync with the formats models actually ship in"), the builder is part of the family, not a follow-up.

3. src/model_list.json points the shipped catalogue at a personal account. https://huggingface.co/vegahyo/Granite-4.2-3B-NPU2 — every other LLM entry is under FastFlowLM/ or Atomic-Germ/. The Hy-MT2 PR that landed immediately before this one added no registry entry at all. Distribution is Atomic-Germ's call; either ask them to host it or drop the model_list.json / model_info.json hunks and let flm-add handle installation, which is what the previous family did.

.opencode/skill/open-granite-kernels/SKILL.md
Cyronius, line 1: should we .gitignore all of our individual agentic stuff?

Cyronius (approved):
a couple of small things but all looks good

PR #22, opened 2026-09-07
perf(attn): remove the per-position scalar float, 20.6x TTFT and 32.1x decode

vegah:
Stacked on #21 (Granite). Review that one first — this branch is 11 commits on
  top of it and touches five kernel files plus `dx.py` and `recipes/dense.py`.

  `OPEN-ATTN-CONTEXT` recorded that a decode step's cost is dominated by a term
  linear in context position — 1.5 ms/position on Qwen3-4B, 2.0 on Granite — so a
  token at position 2048 cost 3–4 seconds. It guessed the mechanism: a fixed
  per-head cost, `attn.h` "not saturating the vector unit at hd 64".

  The direction was right and the mechanism was not. It is **scalar float on the
  scalar unit**, in a loop that runs `heads × positions` times per layer. Two
  ablations on the same build settle it, decode step at position 2048:

  - baseline — 185.3 ms
  - drop q's low bf16 half, which **halves the score MACs** — 185.2 ms
  - drop the single `* 1/sqrt(HD)` beside it — **160.1 ms**

  Halving the arithmetic was free. Deleting one scalar operation was 18% of the
  context term. It also explains the head-count correlation that section found
  real and could not account for: the cost is per head, and it is not arithmetic.

  **What follows from that reading**, in six steps: the softmax exponentials batch
  over heads through the vector unit; `1/sqrt(HD)` folds into q as an exponent
  shift (a power of two at both head dims this path serves, so bit-identical); the
  remaining scalars move into the vector phase that already has every head in a
  register; the heads split across five cores, which is the largest divisor the og
  element layout allows; the head loops unroll, because the cost was
  dependent-chain latency rather than throughput; and four cached rows are
  processed per call so one 32-lane exponential covers the block.

  That last one works *because* narrowing the exponential from 32 lanes to 8
  measured ~1%: vector width is nearly free on this core, instruction count is not.

  **Measured through `flm bench`** — Granite 4.2 3B,
  `utilities/bench-configs/bench-1k.json`, 1005 prompt tokens, two builds
  differing only in these kernels:

  - TTFT: 1216.24 s → **58.93 s** (20.6×)
  - prefill: 0.826 → **17.05 tok/s** (20.6×)
  - decode: 0.415 → **13.33 tok/s** (32.1×)

  Twenty minutes to first token on a thousand-token prompt, down to under one. The
  per-position slope goes 2.289 → 0.0247 ms, 92× flatter.

  The cost model behind those numbers was fitted on the *before* run's prefill
  alone, predicted that run's decode to 0.5% on an independent series, and was
  then used to write down 58 s / 17.3 / 14.3 before the second build existed.
  Measured 58.93 / 17.05 / 13.33.

  **Prefill moves for the same reason decode does**, and that is worth stating:
  prefill costs about what a decode step costs *at that position*, so it is
  quadratic in the prompt length and invisible on a short prompt. Nothing here
  batches a prompt — that is untouched, and it is now the dominant term for any
  document-shaped input.

  **Only Granite takes this path** (`recipes/dense.py` sets `VEXP` for it alone).
  Qwen3-4B rebuilt on these kernels gives byte-identical `insts.bin` for all three
  sets and xclbins differing only in build stamps. The knobs are per-geometry
  rather than per-family, so another family joins by measuring the same way.

  **Also here:** `utilities/bench-configs/` — six configs from 1k to 32k carrying
  the built-in default's own prompt, and a README with the arithmetic for why the
  default 32k config may not finish (~342 h on the unoptimised build, ~4.1 h
  after) plus a two-measurement recipe for predicting it on any model.

  `ATTN_NULL` and `ATTN_ABL` stay in `attn.h` as documented probes. They are
  compile-time and off by default, and they are how the mechanism was found.

  Correctness unchanged throughout: 4-layer slice against the fp64 replica, logits
  corr 0.999998 / 0.999984, same argmax, identical top-5.


Cyronius (approved):
What fable found were small nits -- this all looks good to me and my smarter counterpart. Approved.

Atomic-Germ (approved):

Atomic-Germ:
That's some significant improvement

Cyronius:
> That's some significant improvement

I'm looking into bringing those into the other model families now that vegah has done the hard work. Mentioning so we don't double-work.

Issue #14, opened 2026-09-06
add gguf support


Cyronius:
we should support gguf and treat q4nx as a legacy format

Atomic-Germ:
100% Agree

vegah:
I will look into this.

Atomic-Germ:
@vegah I've given it a few very different tries, but it's definitely doable for one of you.

Cyronius:
This is an extremely high value change. I'm excited to see it! I have some rough stuff in my phlegm codebase but you will probably find that starting over is easier.

Atomic-Germ:
@vegah  Would this be a starting point? #24

Cyronius:
still in progress? anything I can do to help push this along this weekend? it's kind of funny that we all have separate gguf implementations. We need one amazing one, but for now I would settle for one proof of concept that can only handle Q4_0 or Q4_1.

Atomic-Germ:
> still in progress? anything I can do to help push this along this weekend? it's kind of funny that we all have separate gguf implementations. We need one amazing one, but for now I would settle for one proof of concept that can only handle Q4_0 or Q4_1.

Same. Actually I've now tried three alternate ways. And then there's the one amd devs are working on privately. And then there's the llama.cpp effort for npu inference.

This all has to converge, I'm sure it will. I'm definitely struggling to get anything that isn't painfully hacky and the last thing we need is to be self-hampered by needing to support specific models.

I've made most of my progress toward Q4_K and a bit toward mxfp4, but I'm having a hard time not ending up with a near useless translation layer, or some kind of repack.

Still trying though. It'll happen. It's just hard.

Atomic-Germ:
Looks like @chizamd made progress on this: https://github.com/ROCm/FastFlowLM/pull/724/changes#diff-9fa36bc5938fcb1c1dd2dc60fcabbcbc8df1022a718c4089f5c6d9f86f19e266

EDIT: He didn't get very far in actually, it has some huge flaws.

Cyronius:
> Same. Actually I've now tried three alternate ways. And then there's the one amd devs are working on privately. And then there's the llama.cpp effort for npu inference.

[...]

Issue #9, opened 2026-09-05
Update build information


vegah:
The build process is a little more complicated now.
Both the xclbins and the executable needs to be built. 
We need to update the build information so it reflects how to build it on windows and linux from source to a working version. 
Suggestion:
Move build information from readme.md to build.md and link to it instead. build.md should have a step by step guide on how to build the complete project from scratch.

Atomic-Germ:
I think the best thing to do is to automate the hell out of it, and keep it behind a flag.

Cyronius:
A user following BUILD.md on Linux got stuck on a small thing that turned out to sit on top of a bigger one. #53 fixes both, but the bigger one is really this issue's question.

**The small thing:** BUILD.md's Linux section said `docs/setup.sh` installs the system dependencies. It doesn't - that script installs Xcode command line tools, Homebrew, Ruby and Jekyll, then serves the documentation site from a hardcoded `$HOME/oflm/OpenFlowLM`. macOS-only, nothing to do with building. The PR points readers at `linux-getting-started.md` instead, which has the real `apt install` list, and renames the script `serve-docs.sh` so the name stops inviting the same mistake.

**The bigger thing:** there are two presets called `linux-default` and they build different things.

- `CMakePresets.json` at the top of the repo sets `OFLM_BUILD_KERNELS=ON`, so it builds the engine and the NPU kernels together, creating the kernel toolchain if it's missing. README.md documents this one, run from the repo root.
- `src/CMakePresets.json` has no such option. `cd src && cmake --preset linux-default` gets you the engine only. BUILD.md documents this one.

So README says the build produces the kernels and BUILD.md says it can't, and each is right about its own preset. Someone who follows BUILD.md ends up with a binary that refuses to load a model - the exact failure BUILD.md opens by warning about.

`linux-getting-started.md` is a third starting point, and `clean_build.sh` on the staging branch names README.md and linux-getting-started.md as "the documented Linux procedure" without mentioning BUILD.md at all.

#53 describes both paths honestly rather than picking one, because picking one is your call. Two questions before anyone writes more into these files:

1. Which is the build doc? My guess is README.md keeps the quick version and BUILD.md becomes the detailed one, with linux-getting-started.md covering only drivers and XRT - but it's a guess.
2. Does `cd src` stay supported, or does everything move to the root presets?

Once that's settled I'm happy to do the rewrite. I'd rather not guess and collide with the staging work.

Issue #17, opened 2026-09-06
Open engine: the shipped model list offers models it can't read, and the kernels it needs aren't in the repo


Cyronius:
Someone trying to test the open-kernels engine can't get a model to load, and neither of the two reasons is discoverable from the repo. Both are our fault, not theirs. Raised by @Atomic-Germ on #13.

## Problem 1: the model list offers models this engine can't read

The open engine reads one weight format: the older q4_1 container, the one FLM 1.0.2 wrote. Newer FLM builds write a different format (Q4_K), which packs the numbers in differently and needs a different unpacking routine that we haven't written.

Five entries in `src/model_list.json` are the newer format:

- `qwen3.6-moe:35b-a3b` — the headline model of #13
- `qwen3.5:0.8b`, `qwen3.5:2b`, `qwen3.5:4b`, `qwen3.5:9b`

If you `flm pull` any of those today and point the open engine at it, you get an error and nothing else. The 35B is the sore point: it's the model the PR is named after, and it's the one a new person is most likely to reach for.

These entries are the older format and do work: `qwen3:*`, `llama3.1:8b`, `llama3.2:*`, `gemma3:1b`, `gemma3:4b`.

Nothing in the list says which is which. `"quantization_level"` in the registry is the tell, but you'd have to already know that the open engine cares, and nobody told you.

The engine does at least fail honestly rather than producing garbage — [`q4nx_file.cpp:66`](src/open_qwen36/q4nx_file.cpp#L66) checks the format and says so:

```
q4nx: <path> has 4736-byte quant chunks (FLM 1.0.3 / Q4_K?); the open engine
reads the 1.0.2 q4_1 container only
```

**Related trap that can cost you 22 GB:** every registry entry carries a `flm_min_version`. If that's newer than the version stamped in the model file you already have on disk, FLM decides the model is out of date, **deletes** the files that don't match and downloads them again. If you're testing against a hand-built or hand-edited model, edit `flm_min_version` in the staged `out\model_list.json` first. I've lost a 22 GB download to this.

## Problem 2: the kernels aren't in the repo, and nothing says so up front

The compiled NPU kernels are `.gitignore`d on purpose — same rule as the BERT design sets, the source of truth is the design + recipe, not the binary. But the consequence is that a fresh clone has **no** kernels for **any** model. Until you export them, every model fails, including the ones whose weight format is fine.

[...]

PR #24, opened 2026-09-08
Proposed gguf support

Atomic-Germ:
As in #14 & #33 and also #9 tangentially

This isn't complete, but it's getting there


vegah:
Haha, I had a branch almost ready, but let's keep it one branch :)

open_kernels/designs/lm_head_q8/lm_head_q8.h
```diff
@@ -53,6 +74,22 @@ static constexpr unsigned kPerCall = LMHEAD_PER_CALL;
...
 //   y[r] += scale[kb][r] * 2^-s[kb] * part[r]   (bf16 hi/lo split)
 // Runtime kt/first, noinline + inline (COMDAT): one body in program memory.
+#if LMHEAD_SCALES_F32
+// f32 plane -> bf16 lanes, round-to-nearest-even in INTEGER ops ((u + 0x7FFF +
+// ((u>>16) & 1)) >> 16, the same law q4nx_file.hpp's f32_to_bf16 uses). The
+// accum-roundtrip float convert mis-schedules inside this kernel's kb loop.
+static inline aie::vector<bfloat16, kRows> load_scale(const float *__restrict p) {
+  const aie::vector<uint32_t, kRows> w = aie::load_v<kRows>((const uint32_t *)p);
+  const aie::vector<uint32_t, kRows> r =
+      aie::add(w, aie::add(aie::bit_and((uint32_t)0x10000, w), (uint32_t)0x7FFF));
```
vegah, line 91: Claude:
 This is the lm_head_q8 KNOWN ISSUE, and I don't think it's scheduling or register pressure — it's this expression.
  
The comment three lines up states the law correctly: (u + 0x7FFF + ((u>>16) & 1)) >> 16. The ((u>>16) & 1) term is
 the result's LSB as a 0/1. aie::bit_and(0x10000, w) is that bit left in place, so it adds 0x10000 where the law
 wants 1 — which after the >>16 is a whole extra bf16 ULP on every scale whose mantissa LSB is set.

Checked against torch.to(bfloat16) over 400k plausible scale values:

| expression                         | differs | max rel  | mean rel |
|------------------------------------|---------|----------|----------|
| (u + 0x7FFF + ((u>>16)&1)) >> 16   | 0.00%   | 0        | 0        |
| (u + 0x7FFF + (u & 0x10000)) >> 16 | 49.90%  | 7.81e-03 | 5.69e-03 |

  One bf16 ULP is 2^-8 = 3.9e-03, and half the scales are affected — which lands right on the maxrel ~5e-3 / cos 
0.999995 the SKILL.md reports, against the q4nx build's 4.5e-6. Minimal fix, using the same downshift already on the
 next line:
 
 aie::add(w, aie::add(aie::bit_and((uint32_t)1, aie::downshift(w, 16)), (uint32_t)0x7FFF));

 Worth re-running the random-code fixtures after — if this is it, the f32 twin should land at the q4nx build's 4.5e-6
 and unblock the MoE phase.

open_kernels/gguf_pool.py
```diff
@@ -0,0 +1,403 @@
...
+min). Nothing is requantized or narrowed.
+
+Why f32 scales in the pool: XDNA2 has no fp16 vector type, so the kernel could
+only read GGUF's fp16 scales through ~200 extra integer vector ops per chunk
+(per-bit decode of f32) or a 3-mantissa-bit lossy pack-side narrowing (what
+q4nx-build does today). f32 is the exact middle: the pack widens losslessly,
+and the kernel converts f32 -> bf16 with the native one-op RNE convert the
+epilogue already uses for its accumulators -- so a GGUF model computes with
+bit-identical numerics to its q4nx-converted twin (both are RNE-bf16 scales
+inside the MAC), while the file keeps the GGUF values exactly.
```
vegah, line 17: Is the 6144 B chunk buying anything the kernel keeps? This paragraph and gemv_q4.h:138 both say the result is the
same as pre-narrowed bf16 scales — and that's exactly right, which is what makes me ask.

fp16→f32 is exact, so RNE-to-bf16 of the widened f32 is the same bf16 you get narrowing the fp16 at pack time.
Verified over all 63,488 finite fp16 bit patterns: torch.to(bfloat16) (what q4nx-build does) is identical to the
kernel-side RNE on the widened f32, no exceptions.

If that holds, the "3-mantissa-bit lossy pack-side narrowing" isn't a loss relative to this design — the kernel
performs the same narrowing, just later — and the f32 plane costs +20% on the q4 chunk (6144 vs 5120; the q8 head is
+5.9%). On Granite 4.2 3B that's ~2.13 GB → ~2.55 GB of pool. A decode step here measures ~50 ms with ~46 ms in
part0, which at the ~46 GB/s shim roof is essentially pure weight streaming, so I'd estimate ~+8 ms/step (≈19 → ≈17
tok/s). Estimate, not a measurement.

The counter-argument I can see is keeping the file exact for a future kernel that wants more than bf16 of scale —
worth stating explicitly if that's the intent, since it's the whole justification for the *_f32 twins and the
doubled export.

src/open_qwen36/pools.cpp
```diff
@@ -54,14 +58,153 @@ const uint8_t* raw(const Q4nxFile& m, const std::string& name, size_t need, size
...
+        const std::string name = with_layer(op.tensor, layer);
+        if (op.nch == 0 || op.in_dim == 0) fail("std_perm_gguf " + name + " without nch / in_dim");
+        const GgufFile& g = dynamic_cast<const GgufFile&>(m);
+        const GgufFile::TensorInfo& t = g.tensor(name);
+        // Q8_0 / K-quant tensors (typical GGUF lm_head / tied embeddings) are
+        // dequantized per row and re-quantized here to the pool's q4 layout
+        // with fp16 scales -- the same conversion q4nx-build does offline. The
+        // matmul body of the repo stays untouched (Q4_1 etc. go through the
+        // byte-exact path below).
+        const bool kq = t.type == GgufFile::Type::Q8_0 || t.type == GgufFile::Type::Q4_K_S ||
```
vegah, line 308: Claude:
Q4_K doesn't have to go through dequantize→requantize — it can be exact, and the difference is measurable.

A Q4_K super-block factors into per-32 t_j = d·s_j and u_j = dmin·m_j, and both products are exact in float32 (an
11-bit significand × a 6-bit integer needs 17 bits). Since Q4_K dequantizes w = t_j·q − u_j and q4_1 is w = d·q + m,
that's d = t_j, m = −u_j with the 4-bit codes never touched — the same class of operation as Q4_1, not a
requantization. (The factoring is your own unpack_q4_k on q4nx-build's q4_k branch; its _refit_one_side isn't needed
when the target keeps full-width scales.)

Measured on granite-4.2-3b-Q4_K_M, against what the file encodes (gguf.dequantize):

| tensor          | native    | dequant→requant | ratio |
|-----------------|-----------|-----------------|-------|
| blk.0.ffn_up    | 8.663e-05 | 7.200e-04       | 8.31× |
| blk.0.attn_q    | 8.707e-05 | 7.100e-04       | 8.15× |
| blk.20.ffn_down | 1.634e-05 | 1.369e-04       | 8.38× |

cosine 0.999986 vs 0.999104. This matters for #17 specifically — the five unreadable model_list.json entries are
K-quants, so this path is what determines their quality.

Q6_K genuinely can't be relabelled (6-bit codes, per-16 scales), so a fit is unavoidable there; worth labelling it
as the only lossy step rather than grouping it with Q4_K.

src/open_qwen36/pools.cpp
```diff
@@ -54,14 +58,153 @@ const uint8_t* raw(const Q4nxFile& m, const std::string& name, size_t need, size
...
         auto perm = std_perm(op.nch, op.in_dim);
         for (size_t c = 0; c < op.nch; ++c) std::memcpy(dst + op.dst + c * ch, src + perm[c] * ch, ch);
+    } else if (op.op == "std_perm_gguf") {
+        // A GGUF [out, in] matmul tensor (Q4_0 / Q4_1 blocks, row-major, fp16
+        // scales) into the pool's band order of f32-scale chunks (6144 B): the
+        // codes permute into the chunk's 16-lane interleave, the fp16 block
+        // scales widen EXACTLY to f32 (open_kernels/gguf_pool.py). Zero loss.
+        const std::string name = with_layer(op.tensor, layer);
+        if (op.nch == 0 || op.in_dim == 0) fail("std_perm_gguf " + name + " without nch / in_dim");
+        const GgufFile& g = dynamic_cast<const GgufFile&>(m);
```
vegah, line 301: Minor architectural worry: this makes the packing interpreter format-aware, and pools.cpp is the one file whose
byte-exactness is the safety property — mirrored in NumPy by recipes/pack.py and frozen against legacy_pools.py by
test_pack_plan.py. A new std_perm_gguf op plus a downcast means both interpreters now have to be kept in lockstep
across two layouts.

The alternative is to have the source hand back bytes already in the chunk layout, so std_perm stays one op and
pools.cpp never learns what a GGUF is. That's what I did on my branch and it kept pools.cpp to +21/−15 with no
arithmetic touched — happy to show it if useful, though it presumes the single-layout answer in comment 2.

vegah:
Added some comments just based on a Claude review (which again is based on my branch)

Atomic-Germ:
I'm not sure what it will take to get the simplified flow working on windows also, but it probably already does. This wasn't really a deliberate out-of-scope addition, but I really did need it in order to work on the rest along the way:

- Repo-root workflow ( not src) with cmake --preset linux-default → cmake --build --preset linux-default → cmake --install build.
- Full-distribution note: the default Linux preset also compiles the open NPU kernel xclbins, requiring XRT (installed) with ironvenv and third_party/mlir-aie+third_party/Peano auto-created/cloned by the build.
- Presets: linux-default (full), linux-debug (engine-only, FLM_BUILD_KERNELS=OFF), linux-portable (bundled libs), windows-default.
- Kernel subsetting: -DFLM_KERNEL_SPECS=….
- Tests: ctest --preset linux-default (the flm list smoke test).
- Packaging: cpack --preset linux-package-tgz/-deb/-rpm (run from repo root).
- One-shot: cmake --workflow --preset linux-default.
- Bumped the CMake prerequisite to 3.25 (needed for the test/package/workflow presets).

I've also merged main back in since a lot happened since.

Cyronius:
it's all good. I'll try windows soon and get a PR going if it's no bueno.

Cyronius (approved):

vegah:
I am working on fixing the conflicts...

vegah:
(Or, Claude is!)

vegah:
Merged `main` into this branch and pushed the resolution (`f3b9cda`) — it was conflicting in six files.

main had moved under it four times: the FLM→OFLM rename and the preset build (#47), the GEMM-block prefill route (#39), the VLM / M-RoPE path (#45), and a local `lm_head` kernel TU. Each collided with the `WeightFile` seam.

The rule throughout: **take main's side, then route it through the weights view.** `w_` is `man_` for a `.q4nx` and the GGUF-derived manifest for a GGUF, so main's new code reading geometry, tensor names or the tail off `man_` would read the wrong manifest the moment the model is a GGUF — silently, since both are valid manifests. 18 such sites now go through `w_`, and the GEMM-block embed moved from `Q4nxFile::bf16_row` to the `WeightFile::embed_row` virtual (which is what dequantizes a GGUF row).

Two that weren't mechanical:

- **`core.cpp`** keeps this branch's GGUF config derivation (a GGUF needs no `config.json`) beside main's `image_token_id` / `mrope_section` parsing, and its verbose embed check beside main's image-embedding path.
- **`lm_head_q4.py`** keeps main's locally *generated* kernel TU — `designs/layer_x`'s copy is per-spec and gets removed again, which is what broke Gemma's `lm_head` build on 2026-09-08 — and gives it this branch's `GEMV_Q4_PREFIX` wrapping, so the f32-scale build still resolves `gemv_q4s32_gy`. The prefix is a compile flag, so one generated source serves both builds.

`pools.cpp` keeps the source-agnostic `raw(const WeightFile&)`; main's `raw(const Q4nxFile&)` is subsumed. `README.md` and `src/CMakeLists.txt` take main's side wholesale — main had already absorbed this branch's build-from-source rewrite under the new name — except `open_qwen36/gguf_file.cpp`, re-added to the engine sources. `build.cmd` gets it too and now runs `gguf_pack_test`.

Verified on Windows/MSVC against the system XRT: `manifest_test`, `pools_test` (Q4_K transcode and the mixed pools included) and `gguf_pack_test` all PASS, and `open_qwen36_cli` links. Not run: anything on hardware, and the Linux kernel export.

— Claude, on behalf of @vegah

vegah:
@Atomic-Germ Merge if it is okay, if not - just remove the latest commit.

vegah (approved):

src/open_qwen36/pools.cpp
Atomic-Germ: I'm definitely interested in that, at the very least we need to be looking at alternative methods of accomplishing this just in case there's an altogether better way we're missing.

vegah: Happy to. It is branch `gguf-runtime` off `0fe7922` (PR #21), four commits, never pushed — say the word and I will put it up.

**The seam is one header**, `weight_source.hpp`. Three methods — `raw(name)`, `embed_row(...)`, `drop_pages()` — and one contract:

> a source hands back, per tensor, EXACTLY the bytes `model.q4nx` would have held for it, in the file's raster order (chunk `f` covers rows `32*(f/ncol)`, cols `256*(f%ncol)`).

`Q4nxFile` satisfies that by pointing into its mapping; a GGUF source satisfies it by building the bytes. `raw()` returns a value rather than a pointer — `TensorBytes`, which is borrowed / solely owned / shared with a cache and looks the same to the caller — because bytes that had to be built need an owner and only the caller knows how long it needs them.

**`pools.cpp` is then +21/−15, and all of it is two mechanical shapes.** No op is added, `std_perm` stays one op, and no arithmetic is touched:

```diff
-const uint8_t* raw(const Q4nxFile& m, const std::string& name, size_t need, size_t* got = nullptr) {
-    size_t n = 0;
-    const uint8_t* p = m.raw(name, &n);
+TensorBytes raw(const WeightSource& m, const std::string& name, size_t need, size_t* got = nullptr) {
+    TensorBytes b = m.raw(name);
+    const size_t n = b.size();
     if (n < need) fail(name + " is " + std::to_string(n) + " B, the plan needs " + std::to_string(need));
     if (got) *got = n;
-    return p;
+    return b;
 }

-void apply(const PackOp& op, const Q4nxFile& m, int layer, uint8_t* dst, size_t dst_bytes, size_t ch) {
+void apply(const PackOp& op, const WeightSource& m, int layer, uint8_t* dst, size_t dst_bytes, size_t ch) {
     if (op.op == "std_perm") {
         ...
-        const uint8_t* src = raw(m, name, (op.chunk0 + op.nch) * ch) + op.chunk0 * ch;
+        const TensorBytes srcb = raw(m, name, (op.chunk0 + op.nch) * ch);
+        const uint8_t* src = srcb.data() + op.chunk0 * ch;
         auto perm = std_perm(op.nch, op.in_dim);
         for (size_t c = 0; c < op.nch; ++c) std::memcpy(dst + op.dst + c * ch, src + perm[c] * ch, ch);
```

The other five ops (`expert_stripes`, `expert_down`, `put`, `lmhead_q8`, `conv_transpose`) are each the same two lines, and `pack_pool` / `pack_consts` / `pack_lmhead` are signature-only. `pools.hpp` swaps its `q4nx_file.hpp` include for `weight_source.hpp`.

**The format work does not disappear, it moves** — `gguf_repack.cpp` (281 lines: the block formats to raster chunks, Q4_K included), `gguf_source.cpp` (214: the `WeightSource`), `gguf_file.cpp` (364: the container). That is *more* total lines than this PR's. The trade is only about which file they are in: none of them is the one `recipes/pack.py` mirrors and `test_pack_plan.py` freezes.

**Two things I should be straight about:**

- It is based on `0fe7922` and predates the OFLM rename (#47), the GEMM-block prefill route (#39) and the VLM path (#45), so it needs a rebase before it reads as a diff against today's `main`. I have not done that rebase.
- Its byte-exactness evidence is `specs/open-engine/tests/test_gguf_pack.py`. The `test_law_*` half is self-contained — random blocks, no model, no network — and pins the NumPy statement of the law against both `open_kernels/q4_1_pack.py` and `gguf_repack.cpp`, so a change to any one is caught by the other two. The `test_model_*` half needs a real GGUF and **skips** without one; it is not currently run anywhere, and I have not run it against the current tree.

And to be explicit about the dependency I flagged: this shape presumes comment 2's single-layout answer. If the f32-scale plane stays, the source has two output layouts to produce and the seam has to carry which one — still outside `pools.cpp`, but no longer free.

— Claude, on behalf of @vegah

Cyronius:
I was trying to make sure my code was going to agree with this, and opus uncovered something:

I think the Q4_0 path is off by a constant.

A GGUF Q4_0 block stores nibbles 0-15 that stand for -8 to 7, so llama.cpp's
dequant is `(q - 8) * d`. The pack here uses `q * d`: the min plane is zeroed
for Q4_0 (`pools.cpp`, `mf = has_min ? fp16_to_f32(um) : 0.f`), `embed_row`
does `d * (p[2 + j] & 0xF)`, and `gguf_pool.py` fills the min plane with
zeros. Every weight comes out 8*d too high - a block that should straddle zero
ends up entirely on one side of it.

The kernel already computes `q * d + m`, so writing `m = -8 * d` into the min
plane fixes it with no kernel change and no loss: -8 * d only moves an
exponent, exact in f32 and bf16 alike.

Why the tests don't catch it: `gguf_pool`'s reference dequant uses the same
`q * d + 0`, so the packer and the checker agree with each other, and
`gguf_pack_test` only builds Q4_1 and Q8_0 tensors - there's no Q4_0 fixture at
all. We hit the same thing on the .q4nx side last week and it was invisible for
exactly that reason; what finally caught it was the model answering with noise
while every numerical check passed.

One other thing worth flagging while you're in here. "Q4_0" now means two
different byte layouts in this tree. GGUF's nibbles are offset by 8, so 0 means
-8. The .q4nx containers q4nx-build produces hold the same signed values as
two's complement, where 0 means 0 and 8 means -8 - `unpack_q4_0` subtracts the
8, then `model_converter` packs the signed int8 straight back with `& 0x0F`.
Same numbers, different labels, and reading one as the other shifts everything
by 8*d. Probably worth separate names before the two paths meet.

src/open_qwen36/pools.cpp
Atomic-Germ: After way too much thought; you're definitely right.

PR #26, opened 2026-09-08
open kernels: the seventh family -- Qwen3.5 dense, all four sizes

Cyronius:
Stacked on #25 — read that one first. Until it merges, the diff here includes its commits too.

Part of #17.

Qwen3.5 dense runs on the open kernels, in all four published sizes: 9B, 4B, 2B
and 0.8B. This is the biggest single item on the model list — 19 published
models, around 23.5k downloads a month.

Almost none of it is new code. A Qwen3.5 dense layer is a Qwen3.6-MoE layer
with the mixture-of-experts block swapped for an ordinary gated feed-forward
one. Everything else is the 35B's and we already run it. So the new recipe
mostly calls the MoE one for the attention half and the dense one for the
feed-forward half. The kernel design changes under `designs/layer_x` and
`designs/dn_glue` are the feed-forward tail on a layer that was written for a
MoE — they touch files the other families share, and rebuilding those families
on them produces identical output, but this family is why they exist.

**The 4B is the exception, and it is the part worth reading.** Its container
stores the attention output projection at 8-bit precision, like its three
siblings. Running that natively needs a kernel carrying both weight formats,
and at the 4B's width that kernel does not fit in the core's 16 KB of program
memory — with the size optimisations already applied and no levers left.
Before this PR, asking for it succeeded and then failed inside the compiler a
minute later, with an error about program memory that says nothing about the
model you asked for.

So the recipe now records which widths have a working dual-format kernel, and
for a width that does not, it quietly falls back to the older format and prints
one line saying what that cost — measured against the container's own 8-bit
values, the difference is about three parts in ten thousand. It falls back
rather than refusing, because refusing strands the 4B, and someone would rather
have the model that works and be told the price than an error they cannot act
on.

**One thing to know before testing this.** `model_list.json` points the
`qwen3.5:*` entries at the `flm_q4k_high_precision` branch, whose container the
open engine cannot read (that is Problem 1 in #17, and it is not fixed here).
These four sizes run on the container on the repo's default branch. So
`flm pull qwen3.5:9b` followed by the open engine will still fail, and that is
expected.

Granite landed as the sixth family while this was in review, so this is the
seventh. Rebased on main; both families' catalogue entries and spec sections
are kept side by side.

**What I'd like from you:** whether falling back with a warning is the right
call for the 4B, or whether you would rather it refused outright.

Tests: 130 passed, 1 skipped.


Atomic-Germ:
@Cyronius I think it should warn on 4b with a not-implemented message, which is on the hopeful side.

vegah (approved):

Cyronius:
Done — pushed as `9617be8b`. It warns and keeps going, and reads as a missing kernel rather than a limit:

```
open_kernels: warning: q8 (8-bit) weights are not implemented yet for qwen35 hidden 2560
-- the kernel does not fit in program memory. Falling back to q4_1.
```

The code comment records what a fix takes: a mixed core small enough to fit, then adding the width to `MIXED_CORE_FITS`, which is a one-line change once one builds.

Atomic-Germ (approved):

PR #37, opened 2026-09-09
Open engine: read Q4_K containers, and write them (#17)

Cyronius:
This answers the one item left open on #17 ("decide about Q4_K"): the open
engine can now read the weight format the newer FLM builds write, so the five models in `model_list.json` that use it
stop being off-limits — including the 35B this engine is named after.

**The short version.** There are two weight layouts in circulation. The engine
read the older one (q4_1, what FLM 1.0.2 wrote) and refused the newer one (Q4_K,
1.0.3+). It now reads both. Nothing in the kernels changed, and no kernel needs
rebuilding.

That turned out to be much less work than the earlier plan assumed. Q4_K stores
a scale and an offset for each row and each group of 32 columns — the same
grouping the engine's own layout already uses, and in the same order. So loading
a Q4_K model is two multiplications per weight and a byte shuffle, done once
while the weights are being laid out for the NPU. The kernels never see the
difference. The earlier plan expected to teach them a second decoder and worried
about fitting it in the cores' 16 KB of program memory; none of that is needed.

**What it costs.** Weights land 0.48% off in RMS terms, measured across every
Q4_K tensor of a real model. For comparison, the engine already re-quantizes q8
weights down to q4_1 on this same path, and that costs 8%. So this is the more
accurate of the two conversions the packer does, by a wide margin.

**The converter can also write the format**, via `q4nx-build --quant Q4_K`. That
is the part I'd most like a second opinion on — see the question at the end.

**How it was checked.** Thirteen unit tests, run on both sides of the
Python/C++ boundary and required to agree byte for byte, so the reference
implementation and the engine cannot drift apart. Then a real model: a
Qwen3.5-0.8B fine-tune converted twice from one GGUF, once to each format, so
the weights are identical and the format is the only variable. Both load through
`flm serve` on the open kernels and answer coherently under `flm-test`. At
temperature 0 they produce the same answer, word for word, until they pick
different phrasing for the same claim partway through the second sentence.
Speed is the same within noise.

`flm-test --tools` fails 5 of its 6 checks on the Q4_K model — and fails the same
5 identically on the q4_1 twin. That is a 0.8B model that does not emit tool
calls, not a format problem, but it will show up in the CSV so it is worth saying
plainly.

**One unrelated fix rode along.** `configs/qwen3.5_0.8b.json` marked the
`up_proj` weights as q8 while the model Atomic-Germ actually ships stores them as
q4_1, and the other three qwen3.5 sizes never marked it. The kernel exporter
refuses a model whose feed-forward weights are split across two formats, so
*every* 0.8B container this converter produced failed to export — at the old
format exactly as much as at the new one. Unpinning it matches both the shipped
model and the sibling configs. It is one line, and without it the verification
above cannot be run at all.

**What I'd like from you.** The writer is the part I am least sure about. No
model on Hugging Face or in the registry ships Q4_K yet, so the only container it
has ever been tested against is one it produced itself. That is unavoidable for
now — you cannot test a reader without a writer — but it does mean the reader is
checked against our own writer rather than against FLM's. If the struct moves
again in a future FLM, this catches nothing. What it does guarantee is that a
mismatch shows up as a refusal naming the byte width, never as silently wrong
weights.

So: is shipping the writer the right call now, or would you rather it waited for
a real Q4_K model to check against?

Background, in more detail than belongs here:
`specs/open-engine/plans/archive/q4k-containers.md`.

### DIFF
--- a/open_kernels/model/q4nx.py
+++ b/open_kernels/model/q4nx.py
@@ -18,10 +18,11 @@ The chunk format is PER TENSOR. The stock Qwen3.6-35B keeps only its lm_head at
 q8; its fine-tunes (Darwin, Grug, BigBang, Aquila-mini, Ornith 1.5, and
 Atomic-Germ's own NPU2 mirror) pack attention, linear-attention and shared-expert
 projections at q8 and only the routed experts at q4_1; Qwen3.5 dense containers
-store `ssm_out_proj` and alpha / beta at q8. So nothing is refused at open: each
-tensor is classified from its own shape, and `dq_tile` reads either format. A
-chunk size that is neither 5120 nor 8704 is refused when that tensor is read,
-naming it (Q4_K is 4736; 1280 / 2560 are a smaller chunk geometry).
+store `ssm_out_proj` and alpha / beta at q8; a container written by FLM 1.0.3+
+holds Q4_K (4736 B, `dq_chunks_q4_k`). So nothing is refused at open: each tensor
+is classified from its own shape, and `dq_tile` reads all three. A chunk size that
+is none of them is refused when that tensor is read, naming it (1280 / 2560 are a
+smaller chunk geometry).
 
 Which reading a q8 tensor gets is the packer's decision, mirrored here so a slice
 comparison always measures the KERNELS and never the weights:
@@ -29,8 +30,10 @@ comparison always measures the KERNELS and never the weights:
   * `native_q8(name)` -- true for the projections this kernel set streams at q8
     (the plan's `q8_perm` ops, OPEN-QUANT-Q8). Those read as the container's own
     q8 values, because that is exactly what the pool holds.
-  * `requant_q8` (default True) -- every other q8 tensor reads as the q4_1 the
-    packer writes for it (the re-quantising fallback, OPEN-PACK-PLAN).
+  * `requant_q8` (default True) -- every other q8 tensor, and every Q4_K tensor,
+    reads as the q4_1 the packer writes for it (the re-quantising fallback and the
+    Q4_K transcode, OPEN-PACK-PLAN / OPEN-QUANT-Q4K). Turning it off reads the
+    container's own values, and for Q4_K that difference IS the transcode's cost.
 
 `make_decode.py --requant` forces the whole run -- spec, plan, pools and this
 reader -- onto the fallback, which is the A/B against the q8 path.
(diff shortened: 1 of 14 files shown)


Cyronius:
accidentally closed the issue before i landed this PR, but it goes to #17

vegah:
Two data points from a parallel derivation, plus an answer to the question at
the end.

## Your 0.48% checks out, and I think it can be framed better than you framed it

I reached the same factoring separately — `t_j = d·s_j`, `u_j = dmin·m_j`,
**exact in float32** (an 11-bit significand times a 6-bit integer needs 17 bits),
so the 4-bit codes are never touched — and measured the residual on three real
Granite tensors against what the file encodes (`gguf.dequantize`):

| tensor | native | dequantize→requantize | ratio |
|---|---|---|---|
| `blk.0.ffn_up` | 8.663e-05 | 7.200e-04 | **8.31×** |
| `blk.0.attn_q` | 8.707e-05 | 7.100e-04 | **8.15×** |
| `blk.20.ffn_down` | 1.634e-05 | 1.369e-04 | **8.38×** |

cosine **0.999986** against 0.999104. That cosine implies a relative error of
about **0.53%**, which lands on your 0.48% from a different model — two
independent implementations agreeing to within ~10%.

**The sharpening: that residual _is_ the bf16 scale narrowing, not a Q4_K
conversion cost.** The container stores `d`/`m` as bf16 either way, so a q4_1
container pays the same rounding on its own scales. So it is not "Q4_K costs
0.48%", it is "the container format costs 0.48%, and Q4_K adds nothing on top of
it." Worth saying explicitly, because as written it reads as a price for the new
path — and the 8× against dequantize→requantize is the number that actually
argues for the approach you took.

## On your question — the writer

I think the circularity is more escapable than you allow, and the way out is in
your own PR body: **five entries in `model_list.json` are already Q4_K, written
by FLM 1.0.3+.** Those are real containers from the reference writer,
downloadable today. So the *reader* can be validated against FLM's own output
without waiting for anything, independent of whether the writer ships.

That seems worth doing before merge regardless of the writer decision, since the
reader is the half that silently affects every user of those five models — and
it converts "checked against our own writer" into "checked against the format's
producer", which is the claim you actually want.

On the writer itself: the property you name — a mismatch surfacing as a refusal
that names the byte width, never as silently wrong weights — is the one that
matters here, and it is on the right side of the fail-loud line. I would ship it.

## One coordination note

This touches `src/open_qwen36/pools.cpp` and `pools.hpp`, which a local
GGUF-reader branch of mine also modifies (a `WeightSource` seam, so the
container stops being the format). Not a conflict yet — flagging it so we do not
both rewrite the same function.

vegah (approved):

Atomic-Germ:
I think we should have q4_k support for, if no other reason, ggufs! I've got all embedding models as gguf running on our open kernels, very successfully. So, full gguf compatibility is very near. Even if we don't have the q4_k q4nx's

Issue #32, opened 2026-09-08
Open kernels: batched prefill -- amortize the weight stream across a token block


vegah:
Splitting **item 1 of #16** out into its own issue so it has one owner and
nobody starts it twice. #16 stays as the umbrella for the other two.

Before filing I checked everything open and everything merged since: #20
(linux build script), #24 (gguf), #25/#26/#27/#28 (the shapes → Qwen3.5 →
native q8 → adapters stack), #31, and the merged #21/#22/#23. None of them
touches prefill batching. #23's `max_prefill_len` change is a guard against
`log2` of a non-positive length, not a batch path. The four branches on this
repo are `main`, `staging`, `linux-clean-build` and `proposed-gguf-support`.
So this is unclaimed work.

## Where it stands

Prompt tokens go through the decode step one at a time. From
`src/open_qwen36/README.md` on `main`:

| model | prefill |
|---|---:|
| Qwen3.6-35B-A3B | 124 ms/token |
| Qwen3-4B | 129 ms/token |
| Llama-3.1-8B | 125 ms/token |
| Gemma3-4B | 63 ms/token |
| Granite 4.2 3B (after #22) | 58.6 ms/token |

#22 also established the *shape* of the cost, which matters more than the
column above: **prefill costs about what a decode step costs at that position,
so it is quadratic in the prompt length.** On a short prompt that reads as a
flat per-token rate and hides completely. On Granite it is 58.93 s of TTFT for
a 1005-token prompt — and that is *after* a 20.6× improvement.

For scale, the closed FastFlowLM engine on the same SoC sustains ~320–350
tok/s of prefill (~3 ms/token) on a 9B, and stays flat from 1k to 32k context
(the `flm bench` dump in ROCm/FastFlowLM#636). The open path is roughly 40×
behind on the number a user feels first, and the gap is structural rather than
a tuning deficit: every prompt token re-streams the model's whole weight set.

## The experiment #16 is gated on, worked out on paper first

#16 phrases the open question as whether the q4_1 GEMV dataflow in
`open_kernels/designs/layer_x/` can become a GEMM that amortizes the weight
stream across a token block. The arithmetic says yes, with a lot of room.

- q4nx packs 8192 weights into a 5120 B chunk (512 B bf16 `d`, 512 B bf16 `m`,
  4096 B nibbles) = **0.625 byte per weight**. A projection at one token
  therefore runs at 2 / 0.625 = **3.2 FLOP/byte**.
[...]

PR #38, opened 2026-09-10
Open engine: refuse an out-of-range embed token, and unblock the standalone CLI build

vegah:
Two independent fixes found while working on #32.

**`Q4NX.embed()` had no bounds check.** A token id past the end of
`model.embed_tokens.weight` read whatever bytes follow the tensor in the
container and returned them as an embedding — no error, plausible-looking
values. It now refuses, naming the id and the table size.

**`build.cmd` was missing `DISABLE_ABI_CHECK=1`**, which `CMakeLists.txt`'s
`flm` target already passes. Without it `xrt/detail/abi.h` wants a generated
`version-slim.h` that a raw XRT source checkout never produces, and `cl` fails
with C1083 before reaching any of this file's own code.

### DIFF
--- a/open_kernels/model/q4nx.py
+++ b/open_kernels/model/q4nx.py
@@ -197,14 +197,23 @@ class Q4NX:
         return np.frombuffer(self.raw(name), dtype=np.float32).reshape(t["shape"])
 
     def embed(self, token, hidden=2048):
-        vocab_size = self.tensors["model.embed_tokens.weight"]["shape"][0]
-        # A token id past the end of the table used to read whatever bytes
-        # follow the tensor in the container -- silently, returning plausible
-        # garbage rather than an error.
-        assert 0 <= token < vocab_size, (
-            f"Q4NX.embed: token {token} is out of range for a {vocab_size}-row "
-            f"embedding table (model.embed_tokens.weight) -- valid ids are "
-            f"0..{vocab_size - 1}")
+        # A token id past the end of the table, or a `hidden` that disagrees
+        # with the table's own row width, used to read whatever bytes follow
+        # the tensor in the container -- silently, returning plausible garbage
+        # rather than an error. Raise rather than assert: an assert disappears
+        # under `python -O` / PYTHONOPTIMIZE, which is exactly when a silent
+        # out-of-bounds read is least welcome.
+        vocab_size, row = self.tensors["model.embed_tokens.weight"]["shape"]
+        if not 0 <= token < vocab_size:
+            raise IndexError(
+                f"Q4NX.embed: token {token} is out of range for a {vocab_size}-row "
+                f"embedding table (model.embed_tokens.weight) -- valid ids are "
+                f"0..{vocab_size - 1}")
+        if hidden != row:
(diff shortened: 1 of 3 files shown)


Atomic-Germ:
Can you check this again #47 (Full CMake Build system) so we don't conflict? I believe I caught this in mine but it could be windows specific.

vegah:
Checked against #47 — you did catch it, for CMake: `src/CMakeLists.txt:490` and the three test `CMakeLists.txt` all define `DISABLE_ABI_CHECK=1`, so every CMake-driven build is covered.

`src/open_qwen36/build.cmd` is the gap. It is not a CMake build — four raw `cl` invocations — and none of them carries the flag. Windows-specific, as you guessed, since it is a `.cmd`.

Reproduced on `main` today against a raw `C:\dev\XRT` checkout (no XRT build run in it, so the generated `version-slim.h` does not exist), compiling `core.cpp` with the exact flags `build.cmd` uses:

```
without the flag:  fatal error C1083: Cannot open include file:
                   'xrt/detail/version-slim.h'   [xrt/detail/abi.h(8)]
with the flag:     compiles OK
```

`abi.h` line 7 is `#if !defined(DISABLE_ABI_CHECK)`, so the include is skipped entirely when it is set.

No conflict with #47: this PR touches only the fourth `cl` line, the CLI — the one invocation of the four that takes `/I "%XRT_INCLUDE_DIR%"`. The other three build unit tests that need no XRT, as the script's own comment already says.

The Copilot point on `q4nx.py` is addressed in the branch as well: the `assert` is now a `raise IndexError` (an assert disappears under `python -O`), and `hidden` is validated against the embedding table's own row width rather than trusted.

— Claude, on behalf of @vegah

vegah:
Fair point, and the claim behind it was right: nothing called `Q4NX.embed()` at all. Its only caller is `open_kernels/model/make_decode.py`, so the guard could have been removed or inverted without a single test failing.

Added `specs/open-engine/tests/test_q4nx_embed.py` — nine cases on a container the test builds itself (eight bytes of header length, the safetensors-shaped JSON, then the rows), so no model file is needed:

- every row reads back exactly what was written, through the bf16 round trip
- **both boundaries**: `0` and `vocab_size - 1` must *work* — that is what a guard written `0 < token` or `token <= vocab_size` gets wrong while still passing the middle of the range
- `-1`, `vocab_size` and `vocab_size + 1000` must raise `IndexError`, and the message must name the range
- a `hidden` that is not the row width must raise `ValueError`, including the `hidden=2048` default on a table of another width — the case the token check alone does not catch, since it reads a correctly shaped vector from the wrong offset for a valid token
- the guards fire before any read, checked against a deliberately truncated container

The tests were verified to **fail against the unguarded version**: swapping in `main`'s `embed()` fails 6 of the 9, and the 3 that still pass are the happy-path ones that should pass either way. A regression test nobody has seen fail is not evidence that it would catch anything.

Full spec suite: 192 passed, 1 skipped.

— Claude, on behalf of @vegah

Atomic-Germ (approved):

PR #39, opened 2026-09-10
Open engine: batched prefill as a whole-array GEMM (1.95x TTFT)

### LINKED ISSUE #32: Open kernels: batched prefill -- amortize the weight stream across a token block

vegah:
Closes #32.

Prefill runs as a batched GEMM instead of one dispatch per token. A block of 256
tokens streams the weights **once** through whole-array bf16 GEMMs with q4_1
dequantised on-core; attention still runs per token, unchanged. Off by default
behind `FLM_OPEN_GEMM_BLOCK=1`, so no existing measurement moves.

**TTFT on a ~1005-token prompt: 63.9 s → 32.7 s (1.95×).** Correctness at full
model depth, four sequences: three of four argmax-exact with 5/5 top-5; the
fourth differs at a genuine near-tied logit pair (margin 0.077, 5–27× tighter
than the other three).

**What it does not do.** Attention is untouched and is now ~71% of what remains.
The block is capped at 256 tokens by the design's `M % 256`, and `M > 4` is
impossible on the older GEVM path for a hardware reason (`aie::mmul<4,8,8>`'s
M=4 is what makes token sharing free), which is why this is a GEMM and not a
wider GEVM. RMSNorm and SwiGLU run on the host — a measured ~21% term. Granite
only.

The standalone `open_qwen36_cli.exe` needs the `build.cmd` fix in #38 to build;
`flm.exe` is unaffected.


Cyronius:
is this one ready? prefill being ungodly slow is holding back testing on my end

Cyronius (approved):

Issue #35, opened 2026-09-09
Open engine: a model-adjacent kernel set silently outranks FLM_OPEN_KERNELS_DIR, so a rebuilt kernel set never runs


vegah:
`Engine::find_kernels()` prefers `<model_dir>/open_kernels` unconditionally over
the environment variables, whenever that directory exists and is complete. So
once a model is installed with a kernel set beside it, **`FLM_OPEN_KERNELS_DIR`
and `FLM_XCLBIN_PATH` are silently ignored**, and any locally rebuilt kernel set
never runs.

## Why it is worth fixing rather than documenting

It fails open. There is no error, no warning, and no line in the log saying
which kernel directory was chosen — the engine loads a valid set and produces
valid output, just not from the kernels you built.

Found while measuring a change to the kernels: an A/B came back at 71.19 s
against 70.39 s, i.e. a clean "no measurable effect". Both runs had in fact
loaded the same stale set and were byte-for-byte the same configuration. The
real difference, once the right kernels were actually loaded, was **1.7×**. A
null result that is really "the experiment did not run" is the expensive kind,
because it terminates an investigation instead of prompting one.

## Reproduce

1. Install a model so that `<model_dir>/open_kernels` exists and is complete.
2. Build a modified kernel set somewhere else.
3. Point `FLM_OPEN_KERNELS_DIR` (or `FLM_XCLBIN_PATH`) at it and run.
4. The model-adjacent set is used. Nothing says so.

## Suggested fix

Two parts, either useful alone:

- **Order:** an explicitly-set environment variable should outrank the
  model-adjacent directory. Explicit beats implicit; that is the reason the
  variable exists.
- **Say which:** log the resolved kernel directory once at load, unconditionally.
  This project has a general habit of reporting *the value read* rather than the
  value intended, and this is a place with none of it — a single line would have
  turned the failure above into a five-second diagnosis.

The second is worth doing even if the first is judged too disruptive to change:
the silence is what made this expensive, not the precedence order itself.

Found while working on #32.

Atomic-Germ:
Was this fixed with #40?

vegah:
Yes — #40 added the log line, which was the part that mattered. Closing.

PR #40, opened 2026-09-10
Open engine: say which kernel set was chosen, and why

vegah:
Addresses #35, in part — see the correction at the end.

The engine resolves its kernel set by one of three rules and never says which
won. All three produce valid output, so a set chosen against the user's intent
looks exactly like the right one: an A/B measurement here silently ran identical
kernels twice and read as "no measurable effect", when the real difference was
1.7×.

It now prints the resolved directory and the rule that picked it:

```
open_qwen36: kernels .../Granite-4.2-3B-NPU2/open_kernels (beside the model)
```

**Correcting the issue.** #35 says the model-adjacent directory outranks the
environment variables. That is true of `FLM_XCLBIN_PATH`, which is only
consulted after it — but **not** of `FLM_OPEN_KERNELS_DIR`, which
`find_kernels()` already checks first and already warns about when it is set and
invalid. I wrote the issue too broadly.

So this changes no precedence. The silence is what made the failure expensive,
and it is the half worth fixing on its own.

### DIFF
--- a/src/open_qwen36/engine.cpp
+++ b/src/open_qwen36/engine.cpp
@@ -25,7 +25,8 @@ namespace open_qwen36 {
 
 namespace fs = std::filesystem;
 
-std::string Engine::find_kernels(const LM_Config& config) {
+std::string Engine::find_kernels(const LM_Config& config, std::string* how) {
+    auto chose = [how](const char* where) { if (how) *how = where; };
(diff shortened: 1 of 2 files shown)


Atomic-Germ (approved):

Atomic-Germ:
I got legitimately confused at "beef120", because my brain read it instead of seeing hex.

Issue #33, opened 2026-09-08
Provide a first version


vegah:
We should make a first version containing exe files, binaries and everything needed to test it. 
As it is now, the build / tooling process is too complicated, but the system should be tested by different users, so we should provide a linux and windows install/zip file that contains everything needed to get started. 
The versioning should reflect that it is a early test version, 0.0.1-beta or something. 

# Issues that _needs_ to be solved before ANY version 
- [ ] #7 - Readme should be updated
- [ ] #8 - Rename .exe file
- [ ] #30 - Add a new folder for openflow
- [ ] #14 - I think this should be done before any release too
- [ ] #17 - We need a simple way to add models

We could add a tag for 0.0.1-beta (or whatever we end up calling it), so it is easier to track it. 

This issue is meant as an initial suggestion, open for discussion.

Atomic-Germ:
My votes are that we should begin with v0.0.1 for the first release -- a first stabilized version is usually 0.1.0 as I understand it. I can't stand that FastFlow just kind of numbers however they feel like; how many times have they had n.n.x (for bug fixes) breaking things? The biggest thing is that gguf support but so far it's going smoothly. And #17 is addressed in #34 . I think we should probably set up a project for it, or milestones at least. Unless nobody will use them. #24 incidentally addresses the build system as well just for qol while making changes, but it's only confirmed for linux. (Fedora at that)

Rename and readme may as well be done lastly, since they are the most public-facing and the readme being in sync with the release is going to be important.

Cyronius:
I agree with breaking out of the versioning that flm uses. I think people are going to be confused on the relationship between the two. As we go forward i think we could consider a product rename as well. At some point it will start being dishonest to call it flm. The gguf is a big deal.

Atomic-Germ:
#47 Now works for building the entire release, oflm and all, and can package it for distribution. All parts from a fresh clone.

Atomic-Germ:
GGUF is the only remaining piece that hasn't been accomplished; I'm certain there's more than one right way so I think we should definitely all try something different and compare.

Atomic-Germ:
The build system now produces an installable RPM package with everything  ready to go with #93

PR #41, opened 2026-09-10
Build: name the binary oflm, not flm

### LINKED ISSUE #8: Rename the .exe-file
The exe-file is currently named "flm". It should be renamed to oflm or something similar.

vegah:
Fixes #8.

The binary is now `oflm` (`oflm.exe` on Windows). A binary named `flm` is
ambiguous on PATH wherever FastFlowLM is also installed, and when the wrong one
answers, its errors read as ours — `flm.exe serve … --embeddingmodel x`
reporting "unrecognised option" from a different program entirely.

The CMake target stays `flm`, so only the file on disk moves and every
`$<TARGET_FILE:flm>` reference is unaffected. The vcpkg build script,
`home_install.sh` and the Inno installer's shortcuts follow it. Directory names
(`share/flm`, `lib/flm`, `~/.flm`) and the `FLM_*` environment variables are
deliberately untouched — those are install layout and contract, not the
executable.

Built and run: `[8/8] Linking CXX executable oflm.exe`, and `oflm --version`
starts.

**What it does not do.** `oflm --version` still prints `FLM v1.0.4`, so a bug
report still cannot say which codebase produced it — the ambiguity raised in the
#8 thread. That is #33.

Name picked from the thread: @Atomic-Germ proposed `oflm`, and @Cyronius called
it "fine as the exe name" while preferring something more distinctive overall.
If the project lands on a different name later this is a one-line change.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -412,6 +412,13 @@ endif()
 
 add_executable(flm ${SOURCES} ${HEADERS})
 
+# The file on disk is `oflm`, not `flm` (#8). A binary named `flm` is
+# ambiguous on PATH wherever FastFlowLM is also installed, and the wrong
+# one answering makes its errors look like ours -- a missing flag rather
+# than a different program. The CMake TARGET stays `flm` so that every
+# reference below is unchanged; only the output name moves.
+set_target_properties(flm PROPERTIES OUTPUT_NAME oflm)
+
 if(WIN32)
     if(VCPKG_TOOLCHAIN)
         # A vcpkg toolchain is active (e.g. the rocm-npu-staging dev.py build or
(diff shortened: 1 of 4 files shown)


Atomic-Germ:
Conflicts are because it's already handled in main; might want to look at the installer parts again though.

vegah:
You're right on both counts — main renamed further than this PR did (the CMake target, the project, the `FLM_*` compile definitions), so the binary rename here is redundant. Closing in favour of #51, which is the installer half you pointed at.

Four things the rename left behind, two of them silent: `FLM_QWEN36_ENGINE` was missed (the README and CMakeLists already document `OFLM_QWEN36_ENGINE`); the legacy user directory stopped being searched because the rename turned that line into a duplicate of the one above it; no environment variable had a legacy fallback, including the installer's own wizard pre-fill; and `flm_min_version` → `oflm_min_version` in the registry data throws out of a function `oflm list` calls per entry, so one pre-rename entry kills the whole listing. Measured: 1 row and an exception before, 40 rows after, with the repo's own registry unchanged at 44.

PR #42, opened 2026-09-10
Docs: a build guide that covers the kernels, not just the executable

### LINKED ISSUE #9: Update build information

vegah:
Fixes #9.

The README's build section described CMake presets and stopped there. It never
mentioned the AIE design sets — which are not checked in, need a second
toolchain (IRON / MLIR-AIE), and without which the binary starts and then
refuses to load a model. That is the failure people actually hit, and nothing
told them it was coming.

[`docs/BUILD.md`](docs/BUILD.md) covers both halves for Windows and Linux, and
the README links to it instead of carrying it. Two traps are called out because
they cost real time here:

- the Windows build needs a **Visual Studio developer environment**; without it
  it fails on a missing `<cstdint>` inside a vendored dependency, naming a
  third-party header rather than the cause;
- the embedding design sets **must not be built concurrently** — the IRON build
  cache is shared and matches on content, so two families sharing a geometry
  delete each other's work.

**Provenance.** The Windows executable build and the embedding-set instructions
were run against this repository. The Linux presets and the LLM kernel export
are transcribed from the build scripts' own documented usage, and BUILD.md says
so rather than implying all of it was verified.

**Depends on #41** for the `oflm` name; if that lands differently, the binary
name in BUILD.md follows it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,58 +140,9 @@ No model rewrites, no tuning — it just works.
 
 ## 🛠️ Building from Source
 
-For developers who want to build FastFlowLM from source, we provide CMake presets for a convenient and consistent build experience.
+**[docs/BUILD.md](docs/BUILD.md)** — a step-by-step guide for Windows and Linux.
 
-### Prerequisites
-
-- Git
-- CMake (version 3.22 or higher)
-- A C++20 compatible compiler (e.g., GCC, Clang, MSVC)
-- Ninja (recommended)
-
-### Build Instructions
-
-More details on the exact procedure, with dependencies to be installed, for Linux can be found in [linux-getting-started.md](docs/linux-getting-started.md).
-
-1.  **Clone the repository:**
-
-    ```bash
-    git clone --recursive https://github.com/ROCm/FastFlowLM.git
-    cd FastFlowLM/src
-    ```
-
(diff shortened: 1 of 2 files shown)


Atomic-Germ:
#47 Covers the full build, needs today's commits from main resolved but it's a one-command build system, no manual steps. It also unintentionally covers #41's rename, though goes further. It's the full build + release system.

vegah:
Closing — this is already in `main`, and the other half would now undo newer work.

`docs/BUILD.md` on `main` is **byte-identical** to the file this PR adds, and the README already links it in two places.

The README half has gone stale in the other direction: this PR *deletes* the "Building from Source" section and replaces it with a one-line pointer, but #47 has since rewritten that section for the root CMake build system. Merging it now would remove documentation that is newer than it.

Nothing here is lost — #9 is covered by what is on `main`.

— Claude, on behalf of @vegah

Issue #30, opened 2026-09-08
Add .oflm user directory, scan new → legacy, and put the user directory in the search chains


vegah:
##  Background

  get_models_directory() (common/utils.cpp:254) resolves the user directory to %USERPROFILE%\.flm on Windows and
  $HOME/.config/flm on Linux. The other three lookups — find_model_list(), find_model_info(), find_xclbin_path() — never
  look there at all. They go executable directory → CWD → ../share/flm → build-time prefix.

  That is why flm-add ends by printing:

  export FLM_CONFIG_PATH="$HOME/.config/flm/model_list.json"
  export FLM_XCLBIN_PATH="$HOME/.config/flm"

  The user-level registry and the user-level xclbins/ symlinks only work because the user wires them in by hand.

##  Proposal

  1. Introduce .oflm as the user directory for new installs, keeping the existing per-platform convention:
     %USERPROFILE%\.oflm on Windows, $HOME/.config/oflm on Linux (the dot belongs on Windows only; .config is already
     hidden).
  2. Scan new → legacy: .oflm first, .flm second, in every lookup. An existing install keeps working with no migration
     step and no copying of multi-GB weights.
  3. Add the user directory to find_model_list(), find_model_info() and find_xclbin_path() — ahead of the bundle, behind
     the explicit env var. This is what makes the env exports optional rather than mandatory, and it is the half that
     flm-add currently works around.
  4. While in there: find_model_info()'s Windows branch stops at the executable directory and never reaches ../share/flm
     or the install prefix, unlike its Linux branch. Probably unintentional.

##  Open question — first-wins or merge?

  The two kinds of lookup want different answers, and picking one for both is a trap:

  - Weights / xclbins/<name> — first-wins per entry is right. Look for <name> under .oflm, then .flm.
  - model_list.json — first-wins loses the user's whole registry the moment an .oflm/model_list.json appears, because a
    registry file is not one entry. A merge (built-ins, then legacy user file, then new user file, later never allowed
    to shadow a built-in tag) keeps both.

  Related: flm-add's register() seeds a missing user registry with a full deep copy of the system registry. That
  snapshot goes stale as soon as the application ships new built-in models, and it becomes actively wrong under a merge.
  If scanning becomes new → legacy, the user file should hold user entries only.

  Migration surface

  The user directory holds models/, history/ (runner/runner.cpp:480), and — for flm-add installs — model_list.json plus
[...]

PR #44, opened 2026-09-10
utils: an .oflm user directory, scanned before the legacy .flm one

vegah:
Addresses #30, in part — see what it defers below.

`get_models_directory()` and `xclbin_roots()` now look in an `.oflm` user
directory before the legacy `.flm` one. An existing install keeps working
untouched: no migration step, no copying of multi-gigabyte weights.

Two shapes are scanned on Windows, because the issue asks for
`%USERPROFILE%\.oflm` while `flm-add`'s own `Path.home()/".config"/<name>` would
put it beside the flm directory. Scanning both costs one stat and means a
directory created either way is found.

**Verified by behaviour, not inspection.** With no `.oflm`, `list` reports
`granite:3b ✅`; with an empty `.oflm` present it reports `⏬` — the new root
won; removing `.oflm` restores the first result exactly. So the new path is
inert when the directory does not exist.

**What it defers, deliberately.** The issue's own open question — first-wins
versus merge for `model_list.json` — is untouched. First-wins is right for
weights and for `xclbins/<name>`, which is what this changes; a registry file is
not one entry, and picking first-wins there would lose a user's whole registry
the moment an `.oflm/model_list.json` appears. That decision belongs in the
issue, not in a PR that quietly makes it.

`find_model_list()` and `find_model_info()` are likewise unchanged for the same
reason.

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -140,6 +140,22 @@ std::string user_flm_directory() {
 #endif
 }
 
+/// The user-level directories for THIS project, newest first (#30). Returned as
+/// a list rather than a single path so that a directory made either way is
+/// found: #30 asks for %USERPROFILE%\.oflm on Windows, while flm-add's own
+/// `Path.home() / ".config" / <name>` would put it beside the flm one. Scanning
+/// both costs one stat.
+std::vector<std::string> user_oflm_directories() {
+    std::vector<std::string> v;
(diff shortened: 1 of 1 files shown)


Cyronius:
we should also find ggufs from other common products (lm studio, ollama, etc etc). In the future.

Cyronius (approved):

PR #51, opened 2026-09-11
Finish the oflm rename: four places where the old name stopped being read

vegah:
The rename to `oflm` is done in main, so #41's original content — renaming the binary — is redundant. This replaces it. What is left are four places where a name changed and nothing reads the old one. Two of them fail silently.

**1. `FLM_QWEN36_ENGINE` was missed.** Every other selector is `OFLM_*`, and `src/open_qwen36/README.md` and `CMakeLists.txt:508` already document `OFLM_QWEN36_ENGINE`. Setting the documented name did nothing — no error, a different engine.

**2. The legacy user directory stopped being searched.** The rename turned `user_flm_directory()` into `user_oflm_directory()`, which is a strict subset of the `user_oflm_directories()` call on the line above — so the line became dead code while its own comment went on promising *"an existing install keeps working with no migration and no copying of multi-gigabyte weights"*. The dead duplicate is also why it was invisible. Removed it; `legacy_flm_directories()` restores the search the comment describes.

**3. No environment variable had a legacy fallback.** An install predating the rename has `FLM_MODEL_PATH` in its registry, and both the binary and the installer's own wizard pre-fill read only `OFLM_MODEL_PATH` — so the wizard defaults to `%USERPROFILE%\.oflm` and points an upgrading user away from a model store that is gigabytes. `getenv_oflm()` reads `OFLM_X` then `FLM_X` and reports once per variable which one it used; `oflm.iss` does the same for both values it pre-fills.

**4. The rename changed a field in the registry data as well.** `flm_min_version` became `oflm_min_version`, read by implicit conversion in `check_model_compatibility()`. A pre-rename registry therefore throws `[json.exception.type_error.302] type must be string, but is null` out of a function `oflm list` calls once per entry — so **one old entry kills the entire listing**, naming neither the model nor the field. It now accepts the legacy field, and treats "neither field present" as no constraint rather than as a reason to abort.

### Measured

On a user registry carrying `flm_min_version`:

| | rows listed | error |
|---|---:|---|
| before | 1 | `type_error.302`, exit 1 |
| after | 40 | none |

The repository's own post-rename registry is unchanged at 44 rows, before and after. Environment-variable behaviour checked in all four combinations: `OFLM_*` only, `FLM_*` only (value used, warned once), both (`OFLM_*` wins), neither.

### Not covered

The uninstaller still deletes only the `OFLM_*` registry values, so a machine that ran a pre-rename installer keeps its `FLM_*` ones behind.

https://claude.ai/code/session_01KCT7r8ebHia4kGAWsaKc3Y

### DIFF
--- a/src/common/AutoModel/automodel.cpp
+++ b/src/common/AutoModel/automodel.cpp
@@ -37,8 +37,10 @@ std::string AutoModel::get_current_model() {
 std::unique_ptr<causal_lm> AutoModel::_shared_select_open_engine(const char* env_var, const std::string& family_label) {
 #ifdef OFLM_USE_OPEN_QWEN36
     const std::string kernels = open_qwen36::Engine::find_kernels(*this->lm_config);
-    const char* sel = std::getenv(env_var);
-    const bool use_open = sel ? std::string(sel) == "open" : !kernels.empty();
+    // getenv_oflm, not getenv: an install that predates the oflm rename still has the
+    // FLM_* name exported, and silently ignoring it would pick a different ENGINE (#41).
+    const std::string sel = utils::getenv_oflm(env_var);
+    const bool use_open = !sel.empty() ? sel == "open" : !kernels.empty();
     if (use_open && kernels.empty())
         throw std::runtime_error(std::string(env_var) + "=open but no open kernels were found for " + this->lm_config->model_name);
     if (!use_open)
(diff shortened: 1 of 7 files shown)


Atomic-Germ (approved):

Atomic-Germ:
I think we should leave the uninstaller as-is -- we haven't released an installer at all at this point, so anyone uninstalling their OFLM probably isn't trying to uninstall FLM.

PR #54, opened 2026-09-11
Open kernels: more attention cores (1.09x on Gemma 3), and Gemma 3 12B

vegah:
Two changes. Gemma 3 12B builds, and Gemma 3 decode gets **1.09×** at context.

## 1. More attention cores

Attention ran on `ACORES = NH / HPO` cores, because a core had to own whole `og` elements and an og element was `KVW` wide. It was `KVW` wide because `of_aout` carried **two** things: the two KV cache rows, which genuinely are that wide, and core 0's og output. Gemma 3 is 8 heads over 4 kv — so og_elems is 2, so **two** attention cores on a fabric with room for six.

Core 0 now has an og fifo like every other core, `of_aout` carries only the KV rows, and an og element is `NHL * HD` — the core's own heads. `attn.h` gets `kOGH = min(kNHL, kHPO)`, and `attn_cores()` takes the largest divisor of `NH` instead of `NH / HPO`.

**Measured on gemma3-4b, interleaved against the 2-core set:**

| position | 2 cores | 4 cores | |
|---:|---:|---:|---:|
| 0 | 49.4 ms | 49.6 ms | 1.00× — no attention work at 0 |
| 1024 | 88.0 | 80.6 | **1.092×** |
| 2048 | 94.2 | 86.4 | **1.090×** |
| 3072 | 93.1 | 86.7 | 1.074× |

Tokens: `818,236743` both ways, **identical**. That is the check that matters — splitting heads over more cores is a partition, so a correct implementation cannot move a bit. Predicted 1.099× from the `ATTN_NULL` split (16.3 ms of arithmetic in a 90.5 ms step, halved); measured 1.090×.

**Every other family is untouched**, by construction and by test: `NHL == HPO` everywhere else, so `kOGH == kHPO` and the kernel compiles what it compiled, and ACORES is unchanged (granite 5; llama3.1, qwen3, hunyuan 4). Granite run at both ends:

| position | old topology | split og | |
|---:|---|---|---|
| 0 | 46.8 ms, `198,198,198` | 47.2 ms, `198,198,198` | match |
| 2048 | 97.9 ms, `17010,2564,23955` | 100.3 ms, same | match |

## 2. Gemma 3 12B

Which unlocks `google/gemma-3-12b-it` and every fine-tune of it, including [NbAiLab's Norwegian borealis-12b](https://huggingface.co/NbAiLab/borealis-12b).

**A per-spec core count.** `hidden 3840` is not a multiple of `BAND_ROWS * 8`, while the model's intermediate (15360), q width (4096) and kv width (2048) all are — so one width that did not divide made the model unbuildable against a constant `LIMITS["n_cols"]`. `cores_for(spec)` takes the largest count every width divides by.

That is a widening, not a compromise, and it is measured: on gemma3-4b, 4 cores against 8 is **1.18× at position 0 and 1.06–1.09× with a context**. The weight stream is DMA-bound and the array is one memory client whatever its width — established for the encoder GEMM, confirmed here for the decoder GEMV. Every spec that divides by `64*8` still gets 8.

**An RB cap that knows NHL.** At head_dim 256 the block kernel's working set does not fit L1 once a core owns more than two heads (`'aie.tile' op allocated buffers exceeded available memory`). RB 1 builds. RB 1 → 2 measured 1.22× of the attention *arithmetic* on granite, and that arithmetic is ~18% of a decode step.

**Four catalogue points**, each a wider number on an existing template: the attn tuple `(256, 16, 8, 256, True, False, False)`, `ln` width 3840, `lm_head_q4` K 3840, `gemv_q4` K 3840 and 15360.

## Not verified

**No Gemma 3 12B container exists on this machine, so those kernels build but have not been run.** Correctness against a reference needs a packed model and is the next step.

## One trap worth passing on

The first attempt at (1) kept the `KVW` element and drained only `NHL * HD * 2` of it. It compiled clean and hung on hardware:

```
ERROR: open_qwen36: kernel dx_local at position 0 ended in ERT state 8 (timeout)
```

The element size and the drain size have to agree — the same shape as the documented case where a design's outer I/O signature disagrees with what the host allocated, one layer further in.

https://claude.ai/code/session_01KCT7r8ebHia4kGAWsaKc3Y

— Claude, on behalf of @vegah


Atomic-Germ:
Awesome!!

Cyronius (approved):
hell yeah

PR #66, opened 2026-09-12
Embeddings: oflm bench-embed, /v1/embeddings batches, and the tests for both

### LINKED ISSUE #70: POST {} segfaults the server on /api/show, /api/generate and /v1/completions
Four handlers read a request field through an unchecked `const json&` `operator[]` as one of their first statements. Reading a missing key that way is undefined behaviour, and on this build it segfaults rather than throwing.

**Three of them kill the server process with a one-line request.**

## Measured

v0.1.0 dev, commit `743bc39`, Windows, MSVC 14.51. One server lifetime per endpoint, because a crash ends the run. `POST` with the body `{}` and `content-type: application/json`:

| endpoint | `POST {}` | server afterwards | unchecked field |
|---|---|---|---|
| `/api/show` | **segfault** | dead | `model` |
| `/api/generate` | **segfault** | dead | `prompt` |
| `/v1/completions` | **segfault** | dead | `prompt` |
| `/v1/audio/transcriptions` | 500 | alive | `model`, `file` |

The three crashes are `Segmentation fault` on the `oflm serve` process, confirmed by the shell that started it. The audio handler survives with a 500 — its second read is `request["file"].get<std::string>()`, which throws before the first one's UB has consequences.

The lines, in `src/server/rest_handler.cpp`:

```
 687  handle_show                          std::string model  = request["model"];
 725  handle_generate                      std::string prompt = request["prompt"];
1707  handle_openai_completion             std::string prompt = request["prompt"];
1643  handle_openai_audio_transcriptions   std::string model  = request["model"];
1644                                       std::string file_content = request["file"].get<std::string>();
```

## Why it is worse than a 500

The embeddings handler had the identical defect and it has just been fixed in #66. There, the process went down **after logging `NPU Locked!`** — i.e. holding the NPU access lock. Its own last lines:

```
[LOG]  Body:
{}
[🟢 ]  NPU Locked!
<process gone>
```

`server.cpp:132-150` already documents a related incident: a malformed body once wedged the server permanently because the *error handler* threw on the bytes it was reporting, leaving that same lock held. This is the same resource reached by a different route.

## Expected

`POST {}` to any of these should be a **400** naming the missing field, with the OpenAI-shaped error object the validated paths now use (`message` / `type` / `param` / `code`). Nothing should reach a field read before the body has been checked.

## The shape of the fix, already applied once

`handle_embeddings` in #66 validates in the order the failures occur — body is an object, the required field is present, then every field through a checked accessor — and its patch deliberately anchors on the function signature so it could not touch these three by accident. Same pattern applies here; the required field differs per handler (`prompt` for generate and completions, `model`/`file` for audio, `model` for show).

Worth deciding per handler whether the field is genuinely required: on `/v1/embeddings` `model` turned out to be **optional** by design (the mismatch check reads `!model.empty() && ...`, so absent means "whatever is loaded"), and only the presence and type checks were missing. Do not assume the others are required without reading what consumes them.

## Related, lesser, and untested

`configure_chat_engine_parameters` (`rest_handler.cpp:574-611`) reads `temperature`, `top_p`, `top_k`, `min_p`, the three penalties, `think`, `reasoning_effort` and `image-max-tokens` behind `request.contains(...)` — so presence is checked, but the **type** is not. `{"temperature": "hot"}` should throw out of the implicit conversion into the generic catch, i.e. a 500 for a client error. Not tested; noted because it is the same class one step milder.

vegah:
Main point of PR:
Add a embedding benchmark, mostly to be able to track progress of embedding kernels over time.    
Noticed that the embedding endpoint did not use batching internally, so the endpoint handlet batching as a for loop. Fixed, now uses embed_batch(), which drastically increases embedding speed of batches.  
Added tests for the embedding benchmark.  
Added a document that contains the embedding benchmarks.  Please see that document for the results of the embedding benchmark. 

Found different bugs, or small annoyances, when implementing:  
Field "input" to embeddings was not validated, returned 200 - now returns 400 if it is not a string.  
If batching input had one element that was not a string, the server gave 500 - now also returns 400 and information about the element that is wrong.  
If no input field was defined, it treated it as empty data and returned 200. Now returns 400.  
If model was set to null, it returned 200 and used the model that was loaded. This also now gives 400.  
If no embedding is enabled, it returns 400 if you try to get an embedding.  
  
All endpoints now refuse POST {} instead of crashing the server.  Fixed #70.
Fixed request_id that was not a string lead to the npu being locked for other requests.

Added request validation in server api spec. Added test script and modified test_embed_task_prompt.py.  
  
--host is now refused by run, pull, remove, check and bench-embed

### DIFF
--- a/docs/docs/benchmarks/embeddings_results.md
+++ b/docs/docs/benchmarks/embeddings_results.md
@@ -37,14 +37,14 @@ other pages at all —
 
 Each stage therefore times **two paths over the same texts**: one
 `embed_batch()` call, and the same texts one `embed()` call at a time — which
-is what `/v1/embeddings` does per input. `Speedup` is the ratio.
+is what a caller doing one request per text gets. `Speedup` is the ratio.
 
 ---
 
 ### **Test System:**
 
 AMD Ryzen™ AI 9 HX 370 (Strix Point) with Radeon 890M, 32 GB DRAM. Mains power,
-NPU otherwise idle, models run one at a time.
+NPU otherwise idle, models run one at a time in one session.
 
 | | |
 |---|---|
@@ -55,7 +55,8 @@ NPU otherwise idle, models run one at a time.
 
 **These are throughput and latency figures for the whole pipeline, and NOT NPU
 kernel claims.** The array is shared, so a wall-clock reading measures how busy
-the machine was as much as how good the kernels are.
+the machine was as much as how good the kernels are — see *A note on how easy
+this is to get wrong* at the end of this page, which is not a formality.
 
 ---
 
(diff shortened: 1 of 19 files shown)


Atomic-Germ (approved):
This is why this project benefits so much from being actually open

Cyronius (changes requested):
Fix the crash in the two other handlers. The same missing check that let an empty request kill the server is still at the top of the model-info handler and the audio one. It's the same three lines each. Leaving them means the PR's main claim only half holds.

Fix or reword the vector-width check. Right now it only catches a mismatch when the numbers don't divide cleanly. If a backend returned half the vectors asked for, it would divide fine and hand back wrong answers. Either compare against the model's actual width, or change the comment so it stops promising more than it does.

Delete the dead check in the benchmark. The null test after picking a backend can never fire, and its error message never prints. It reads as protection that isn't there.

Decide what to do about the docs checker. It needs the saved output of the benchmark run to work, and that file isn't in the PR, so nobody else can run it. Commit the output or say in the file that it's for the author only.

Make the batch-of-1 comparison honest. The batched path does a little extra counting work the one-at-a-time path doesn't, which is likely why one model shows the "slow" path coming out faster. Either drop the extra work from the timing or stop calling the comparison like-for-like in the printed footer.

Two smaller ones: --host is now rejected on commands that used to accept and ignore it, so scripts can break — that needs a note. And when no embedding model is loaded the endpoint still returns success with an empty body.

Separately, the PR description needs cutting before anyone reads it.

vegah:
> Fix the crash in the two other handlers. The same missing check that let an empty request kill the server is still at the top of the model-info handler and the audio one. It's the same three lines each. Leaving them means the PR's main claim only half holds.
> 
> Fix or reword the vector-width check. Right now it only catches a mismatch when the numbers don't divide cleanly. If a backend returned half the vectors asked for, it would divide fine and hand back wrong answers. Either compare against the model's actual width, or change the comment so it stops promising more than it does.
> 
> Delete the dead check in the benchmark. The null test after picking a backend can never fire, and its error message never prints. It reads as protection that isn't there.
> 
> Decide what to do about the docs checker. It needs the saved output of the benchmark run to work, and that file isn't in the PR, so nobody else can run it. Commit the output or say in the file that it's for the author only.
> 
> Make the batch-of-1 comparison honest. The batched path does a little extra counting work the one-at-a-time path doesn't, which is likely why one model shows the "slow" path coming out faster. Either drop the extra work from the timing or stop calling the comparison like-for-like in the printed footer.
> 
> Two smaller ones: --host is now rejected on commands that used to accept and ignore it, so scripts can break — that needs a note. And when no embedding model is loaded the endpoint still returns success with an empty body.
> 
> Separately, the PR description needs cutting before anyone reads it.

Agreed on all points. 
I will try to write the pr texts myself from now on :)  
See updated branch, @Cyronius .

Cyronius (approved):

Issue #76, opened 2026-09-13
enhance oflm test so it actually catches things?


Cyronius:
Should we add better tests to oflm test so that it actually catches things? The ones inherited from flm never catch anything.

A few from the last two weeks:

Asking for a model that doesn't exist got you a different model's answer, with the name you asked for echoed back (#52)
Every error response came back as HTTP 200 with a JSON parser message as the body, so the real error text never reached a client (#52)
Every document sent to /v1/embeddings was embedded as a query. Right shape, right length, deterministic, wrong (#52)
An answer cut off at max_tokens said it had finished normally (#52)
Every tool result arrived four tokens short, so a ticket code came back missing its last character and 18 degrees was reported as 15 (#59)
One client asking for low reasoning effort turned thinking on for everybody else (#59)

vegah:
100% totally agree.

PR #77, opened 2026-09-13
oflm-test: make a failed check fail the run, and add checks for the server

Cyronius:
For #76.

**The tool couldn't fail.** That turned out to be the main reason it never
caught anything: a failed check was written into a column of a CSV file and the
command still exited 0, so nothing — a script, a person glancing at the
terminal — ever saw it. It now prints a summary when it finishes and exits 1 if
any check failed.

The other reason is that the checks it had were about the model rather than the
server. Whether a vision model recognises a seagull tells you about the weights.
Every bug in #52 and #59 was in the code around the weights — status codes,
which model actually answered, whether an answer got cut off — and every one of
them came back as a normal-looking success with a sensible answer in it. So
there's a new suite that looks at those instead, `oflm-test --api`. It checks
that a request the server refuses actually comes back refused, rather than as a
success with the error message buried in the body; that asking for a model that
isn't loaded gets you an error instead of a different model's answer; that an
answer cut short at the token limit says so; that one client's settings don't
carry over into the next client's request; and that the streaming and
non-streaming paths say the same thing about the same request.

Two more embedding checks: that a document isn't quietly embedded as a query,
and that a task name the server doesn't recognise is refused rather than swapped
for the default. And the LLM suite had no checks in it at all — it sent two
prompts and wrote the answers down — so it has some now, and its non-streaming
half is no longer commented out.

**One thing worth flagging, because it's the same trap.** All of these pass
against the current build, since the bugs they're about are already fixed. A
check that has never once gone red isn't worth much. So every new check also has
a unit test that hands it the broken response from before the fix and requires
it to fail. Those need no server and no NPU.

The requirements behind all of it are in a new `specs/server-api/`, since the
#52 and #59 fixes shipped without any.

### Three things I'd like your call on

**The exit code.** It now exits 1 when a check fails, where before it always
exited 0. If you have anything that runs it and expects a 0, this will break it.
There's an `--exit-zero` flag that gives you the old behaviour, but if you'd
rather have it the other way round — always exit 0 unless you ask for the
failure — say so and I'll flip it.

**Whether `--api` belongs in `--all`.** I put it there. It's fast and only needs
a chat model loaded, so it seemed to fit, but it is a different kind of test
from the others and I'm happy to leave it out if you'd rather run it on its own.

**Where the requirements should live.** I made a new `specs/server-api/` folder
because the server's HTTP behaviour didn't have one. If you'd rather it sat
under something that already exists, tell me where and I'll move it.


Cyronius:
I might remove the specs -- this is an artifact of my spec-driven development on my machine. a spec is required for all behavior changes, and tests go in first. let me know what you think of that.

Atomic-Germ (approved):
I like it, definitely more helpful. It was initially developed while I was primarily trying to make unusual models work, that's how I found out most panic "////////" states are configuration and not baked in problems in a model, etc. This is now useful for our whole scope.

Thanks! 

Note: We all have our own development process as well as a shared gestalt of them. I like your spec driven methods, and I use a little bit of a different flow. That's a good thing.

PR #34, opened 2026-09-08
rewrite flm-add in cpp as `flm add`

Atomic-Germ:
This addresses #29 ; needs extensive testing on different systems -- not sure if flm pull should be outright replaced yet. This will need quite a bit done when gguf hits but that's alright.

```
./flm add Atomic-Germ/Ornith1.5-Heretic-9B-NPU2 --family qwen3.5
[FLM]  Using custom model list path: /home/atomic-germ/.config/flm/model_list.json
[INFO] Downloading model files from Hugging Face: Atomic-Germ/Ornith1.5-Heretic-9B-NPU2
[INFO] Downloading config.json...
[FLM]  Download completed: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/config.json
[INFO] Downloading model.q4nx...
[FLM]  Download completed: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/model.q4nx
[INFO] Downloading tokenizer.json...
[FLM]  Download completed: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/tokenizer.json
[INFO] Downloading tokenizer_config.json...
[FLM]  Download completed: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/tokenizer_config.json
[INFO] Downloading chat_template.jinja...
[FLM]  Download completed: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/chat_template.jinja
[INFO] Downloading vision_weight.q4nx...
[FLM]  Download completed: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/vision_weight.q4nx
[INFO] Registered tag 'ornith1.5-heretic:9b' in /home/atomic-germ/.config/flm/model_list.json
[INFO] Linked open_kernels: /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2/open_kernels -> /home/atomic-germ/.config/flm/xclbins/Qwen3-4B-i1-GGUF/open_kernels

Done: Ornith1.5-Heretic-9B-NPU2 installed to /home/atomic-germ/.config/flm/models/Ornith1.5-Heretic-9B-NPU2
Run:  flm run ornith1.5-heretic:9b   (or: flm serve ornith1.5-heretic:9b)
```

PR #86, opened 2026-09-15
`oflm add` added (supercedes #34)

### LINKED ISSUE #82: oflm-add: `--system-list` is accepted but ignored
## Summary
`oflm-add` fails with "Could not locate the system model_list.json (looked next to `oflm` and in /opt,/usr,/usr/local share/oflm). Pass --system-list." even when `--system-list <path>` is passed.

## Cause
In `utilities/oflm-add/oflm_add/__init__.py`:
- argparse defines `--system-list`
- `find_system_model_list()` takes no argument and only checks the `oflm`-adjacent candidate plus the fixed paths; it is called bare (`system_list = find_system_model_list()`), so the parsed `args.system_list` value is never consulted.

## Repro
Any environment where the auto-detect candidates are absent (e.g. running from a build tree):
```
oflm-add /path/to/model-dir --tag foo:2b --family llama3 --system-list /path/to/src/model_list.json
# -> Could not locate the system model_list.json … Pass --system-list.
```

## Workaround that worked
Place a copy of `model_list.json` next to an `oflm` entry on `PATH` (that is the first auto-detect candidate), e.g. `ln -s <build>/src/oflm /usr/local/bin/oflm && cp src/model_list.json /usr/local/bin/`.

## Suggested fix
`system_list = Path(args.system_list) if args.system_list else find_system_model_list()` — or thread the argument into the finder.

Atomic-Germ:
- Added oflm add <repo> command and help text.
- Supports all existing oflm-add options unchanged.
- Handles relocatable system registry/xclbin discovery.
- Added OFLM_PYTHON and OFLM_ADD_SCRIPT overrides.
- Updated documentation.


Atomic-Germ:
Confirmed with `oflm add Atomic-Germ/Ornith-1.0-35B-A3B-NPU2 --family qwen3.6-moe`, results in usable `oflm run ornith:35b-a3b`

vegah (approved):

Issue #81, opened 2026-09-14
oflm validate passes on machines where oflm run can't open the NPU


Cyronius:
On Linux, `validate` and `run` reach the NPU by two different routes, and
`validate` only checks one of them. It can report a healthy stack on a machine
where every model fails immediately.

A user on Fedora 44 hit this:

```
$ ./oflm run qwen3.6-moe:35b-a3b
Error:  No such device with index '0'

$ xrt-smi examine
...
Device(s) Present
  0 devices found
```

`oflm validate` passed on that same machine — kernel 7.1.13, `/dev/accel/accel0`
with 8 columns, firmware 1.1.2.64, memlock infinity.

Both results are correct. `sanity_check_npu_stack` (`src/src/main.cpp:210`)
opens `/dev/accel/accelN` directly and checks kernel version, DRM version,
firmware version, column count and the memlock limit. Its verdict —
`overall_ok`, `src/src/main.cpp:400` — is built from exactly those five things.
It never touches XRT. But `run` and `serve` open the NPU *through* XRT
(`oflm_rt::device(0)` at `src/runner/runner.cpp:44` and
`src/server/rest_handler.cpp:326`), and XRT can only see an NPU if it can load
its userspace plugin, `libxrt_driver_xdna.so.2`, at runtime. Without that
plugin XRT enumerates zero devices no matter how healthy the kernel side is.

Fedora packages that plugin nowhere, and building XRT from source doesn't
produce it — it comes from the separate `amd/xdna-driver` repo. So a Fedora
user gets a green validate and a runtime failure with nothing connecting them.

Suggested fix: have `validate` construct a device on the same runtime backend
`run` uses, and fold that into `ready`. When it fails, name the missing plugin
as the likely cause instead of passing XRT's `No such device with index '0'`
straight through. An `xrt_ok` field in the `--json` output would keep it
machine-readable.

The docs already cover this, but only as an after-the-fact footnote in three
places (`docs/docs/install_lin.md:213`, `docs/docs/instructions/cli.md:54`,
`docs/linux-getting-started.md:110`). Three copies of "if validate passes but
run fails" suggests validate should just check it.

Happy to send a PR if you want it done this way. Separately: there's no Fedora
section in the install guide at all — I can add one here or on its own.

Atomic-Germ:
[...]

PR #88, opened 2026-09-15
Fedora-specific cmake preset

Atomic-Germ:
Addresses part of #81 regarding Fedora install directions; qol improvement. Fedora wants certain things in certain places.


Cyronius (approved):
i can't test, but looks good to me

PR #90, opened 2026-09-15
open_kernels: batched prefill GEMM (gemm_q4)

### LINKED ISSUE #89: open_kernels has no batched prefill GEMM
`open_kernels/` covers a Qwen3.6-MoE decode step end to end, but every matmul design in it is a GEMV: one weight pass produces one token. That is the right shape for decode, which is DRAM-bandwidth-bound, and the wrong shape for prefill.

Prefill is the compute-bound phase, and it is the only place XDNA2's matrix units can beat an iGPU — but only if one weight pass feeds many tokens. With a GEMV the weights are re-read per token, so prefill runs at decode speed and the NPU's arithmetic throughput is never used.

`designs/gemm_q4_prefill/` is not this. It dequantises to bf16 first, which gives up the integer matrix unit and moves more bytes, and it predates the blocked dataflow work.

What is missing is a design where:

- the weights stream once per pass and M tokens ride along;
- the quantised weights enter the matrix unit *as stored* — no dequantisation step and no offline requantisation. For GGUF that means Q4_K, whose 6-bit sub-scales have to be folded into the integer accumulation instead of being applied in a float epilogue eight times per chunk;
- the activation-side preparation does not eat the gain, which in practice means the host hands the core a table it has already quantised.

There is a further constraint that shapes the design: L1 is the ceiling on M, because the resident accumulator grows with it. Holding one output tile's bands instead of all of them is what buys the larger M.

Filing this so the gap is recorded. A design that closes it is measured and ready to port.

Cyronius:
Closes #89.

Adds `designs/gemm_q4/` — `gemv_q4` with an M dimension, so one weight pass feeds M tokens instead of one. That is the difference between prefill running at decode speed and prefill being compute-bound, which is the only regime where the NPU's matrix units beat the iGPU.

## What makes it pay on weights as they are stored

**Q4_K goes into the matrix unit natively.** Every GGUF 4-bit type scales every 32 weights, which forces the float epilogue eight times per chunk and is what makes the kernel compute-bound. Folding Q4_K's 6-bit sub-scales into the integer accumulation runs that epilogue once per 256-wide superblock instead. No dequantisation pass, no offline requantisation, and the weight stream is 7.5% smaller because Q4_K's metadata is.

**The activation table arrives already quantised.** `prep_host.py` reproduces the core's arithmetic exactly — the bf16 sign-bit-clear exponent trick, the clamped shift, `round_half_even` matching the core's `conv_even` — so the bytes are identical to what the core used to compute, and the core spends nothing on `reduce_max`, the int16 round-and-scale, or the block-sum tree. Correctness is unmoved by the change.

**Output tiling (`GEMM_BPT`) buys M.** L1 is the ceiling, because the resident accumulator grows with M. Holding one output tile's bands instead of all four shrinks it; `GEMM_ACC_DEPTH=1` gives up drain/compute overlap for more of the same. The L1 budget is

    20736 B + M x (576 + 256 x bpt x acc_depth) <= 65536

which predicts every build pass/fail exactly, including which M values fail to fit.

## Measured

Strix, 32 cores, N=8192 K=2048, one session, minima over ten interleaved rounds of 30 submits each (session floor 0.272 ms):

| config | ms | tok/ms | weights KiB/token |
|---|---|---|---|
| untiled, M = 28 | 0.370 | 75.7 | 338 |
| **`GEMM_BPT=2 GEMM_ACC_DEPTH=1`, M = 40** | 0.494 | **80.9** | 237 |
| `GEMM_BPT=1 GEMM_ACC_DEPTH=1`, M = 52 | 0.672 | 77.4 | **182** |

`BPT=2` at M = 40 is the throughput pick. `BPT=1` at M = 52 is the one to use when a decode stream is running concurrently — same wall-clock, 1.86x less weight traffic per token, which is bandwidth the other engine gets to keep.

Verified in this tree through `harness/run_kernel`:

    (cd designs/gemm_q4 && run_kernel run_m40.cfg && python check.py 40)
    ALL PASS  (q4k bar vs ref_q4k.bin: cos > 0.999999, maxrel < 0.005)

at `cos = 0.999999930, maxrel = 4.310e-04`. (The exact maxrel tracks the fixtures `make_q4k.py` derives from `gemv_q4`'s weights, not the kernel.)

## Conventions

Adapted to this repo on the way in, not copied verbatim:

- run.cfg paths are **relative to the design directory**, per `fixture_paths.py`, so one cfg works from any checkout and on either side of the WSL/Windows split.
- No `.bin` fixtures are checked in. `make_q4k.py` re-expresses `gemv_q4`'s weights in Q4_K form and computes the reference; `make_test.py` writes the repacked weights and the cfg; `prep_host.py` writes the activation table.
- Generated `.cc` entry points and `layout_m*.txt` are gitignored the way `dense/`, `layer_x/` and `gemv_q4`'s generated sources already are.

## Reviewing

    cd open_kernels
    GEMM_N=8192 GEMM_K=2048 GEMM_RS=2 GEMM_CORES=32 GEMM_M=40 GEMM_PER_CALL=1 \
      GEMM_BLOCKED=1 GEMM_Q4K=1 GEMM_PREQ=1 GEMM_BPT=2 GEMM_ACC_DEPTH=1 \
      python build_design.py designs/gemm_q4/gemm_q4.py designs/gemm_q4/build_bpt2d1_m40
    cd designs/gemm_q4 && python make_q4k.py && \
      python make_test.py --m 40 --cores 32 --blocked --preq --bpt 2 --build build_bpt2d1_m40

`make_q4k.py` needs `gemv_q4`'s `w_qkv.bin` / `x_qkv.bin` present, the same as that design's own tests.


Atomic-Germ (approved):
Very nice.

Issue #91, opened 2026-09-15
Finish GPT-OSS 20B on the open engine


Cyronius:
GPT-OSS is the last of the five families in #71 that still refuses to build. The
container side is done; what is left is the model side -- a recipe, a packing
step for the experts, and kernels.

**What already works.** The engine reads GPT-OSS's container correctly now: the
expert format is identified as MXFP4 and decoded against OpenAI's own weights,
the half-width weight chunks are placed, a width that would have produced a
corrupt weight pool is refused, and the adapter selects the open engine if a
kernel set is ever installed. None of that produces a token. It is the ground a
recipe stands on, and until last week each of those returned a plausible wrong
answer with nothing raised.

**The packing step for the experts.** The converter fuses gate, up and down into
one tensor per layer; the order is gate and up alternating every 128 rows for
the first two thirds, then down as one block. Offline, and the expert biases
ship twice -- named, and inside each chunk's own padding -- which gives a way to
pin the order by value rather than by inference.

**Three small kernels**, each following the attention-sink precedent: a new
source file or macro under its own flag, defaulting off, with a before/after
object diff across every shipped family as the gate. A clamped SwiGLU for the
experts; a consts slot and one add for the output projection's bias; and the
router widened from its hard-wired expert count to 32 experts and top-4. The
arithmetic for all three is already written and checked against transformers in
the fp64 reference, so this is plumbing plus the diff. Needs the WSL toolchain.

**The long pole is the expert FFN.** GPT-OSS's expert intermediate is the same
width as its hidden. Every family we run has it at a quarter, and that ratio is
spread through four separate places in the hand-placed core layout -- one of
them divides by a count that comes out zero at GPT-OSS's shape. That is a design
change plus new validated points, each needing a hardware compare. Weeks, and it
needs the NPU.

Then the recipe, the fp64 comparison, and `oflm-test --llm` through
`oflm serve`, which is where a model starts working.

The requirements and what each is verified by are in `specs/open-engine/spec.md`
-- OPEN-FAMILY-GPTOSS, OPEN-QUANT-MXFP4, OPEN-PACK-CHUNK-FUSE, OPEN-WIDTH-PAD
and OPEN-EMBED-STRIDE. Happy to send the packer and the small kernels up ahead
of the MoE work if you would rather review them separately.

Issue #71, opened 2026-09-12
Open the next five families — Qwen3-VL needs no new kernel


Cyronius:
Nine model families still need the closed engine. Five of them are worth doing next, and one of those — Qwen3-VL — looks like it needs no new NPU kernel at all.

Qwen3-VL's decoder is plain Qwen3, which already runs on the open kernels. Its vision tower is the one we already wrote: `src/open_qwen36/vision/vit.hpp` was built against Qwen3-VL's own encoder (transformers calls it `Qwen3VLVisionModel`) while opening Qwen3.5 and Qwen3.6, and checked against it to a correlation of 1.00000000. So both halves exist. What is missing is wiring — telling the kernel recipe that a Qwen3-VL config is a Qwen3 config, adding a third family to the engine's image path, and pointing the tower at its config keys.

That was worth confirming before anything else, so I did the offline half:

- A Qwen3-VL config and a plain Qwen3 config of the same geometry now derive the *same* spec hash, which is what makes a Qwen3-VL model link to an existing Qwen3 kernel set instead of building its own. Both container shapes are handled — the decoder nested under `text_config`, and flattened, which is what we ship.
- Qwen2.5 derives too, but the recipe refuses to build it: Qwen2 puts a bias on its query/key/value projections and our attention kernels have nowhere to put one. Better an explicit refusal than a kernel set that quietly drops it.
- 17 new tests, no NPU or model needed. Full suite green.

The rest, roughly in order of how much new kernel work each needs:

| Family | New kernel work |
|---|---|
| Qwen3-VL 4B | none |
| Qwen2.5 3B | bias on the attention projections |
| GPT-OSS 20B | learned attention sinks (expert routing already exists from Qwen3.6) |
| LFM2 1.2B | a short convolution block that replaces attention in some layers |
| Qwen2.5-VL 3B | a windowed vision encoder, plus Qwen2.5 first |

Cyronius:
I'm a big fan of LFM so I'm pretty stoked to get to the point where that one can go in.

Atomic-Germ:
@Cyronius man I read FLM instead of LFM and it read like you got reset 😂

Cyronius:
Four of the five now run on the open kernels -- Qwen2.5-3B, LFM2-1.2B,
Qwen2.5-VL-3B and Qwen3-VL-4B all pass `oflm-test --llm` through `oflm serve`,
[...]

PR #92, opened 2026-09-16
Open engine: four more families, and the groundwork for GPT-OSS

### LINKED ISSUE #82: oflm-add: `--system-list` is accepted but ignored

Cyronius:
Four of the five families in #71 now run on the open engine, and the fifth has
everything underneath a working model in place. It's all in one PR because the
pieces build on each other and upstream squashes on merge. It's large, so what
follows is a map. The other way in is by commit: 61 of the 70 are substantive (the rest are
merges), and each one explains itself.

**What's in it, in the order it was built**

1. **Qwen2.5 on the open engine.** A per-channel offset on the query, key and
   value projections, applied in the attention kernel. And a bug found because
   of it: Qwen2.5's weight file stores 4-bit numbers in a signed form, every
   other model uses an unsigned one, both take the same space, and nothing in
   the file says which. The packer read one as the other, and nothing caught it
   because the reference we check against unpacks with the same function.
   `open_kernels/recipes/dense.py`, `designs/attn/`, `recipes/pack.py`.

2. **`oflm add` fixes.** It looked for the model registry next to a program
   called `oflm`; the release installs as `flm`. And `--system-list`, the flag
   its own error message points you to, was parsed and never used. That's #82,
   closed on the assumption #86 covered it — it didn't, and this does. A
   Windows install could also die part-way on link permissions.
   `utilities/oflm-add/`.

3. **The faster attention path for Qwen2.5.** Per-token cost stops growing with
   conversation length: 19 times faster at 2048 tokens of context. Four
   reference implementations ride along here — LFM2's convolution arithmetic,
   Qwen2.5-VL's windowed tower, GPT-OSS's attention sink behind a flag that's
   off for every existing family, Qwen3-VL's adapter. None changes a running
   model. `recipes/attnknobs.py`, `open_kernels/model/replica_*.py`.

4. **LFM2.** Ten of its sixteen layers replace attention with a short
   convolution, so it's a new layer type, kernel and recipe. Matches the
   reference and answers correctly. On the faster path its convolution layers
   cost the same at any conversation length; the six attention layers account
   for all of the growth. `designs/short_conv/`, `recipes/lfm2.py`.

5. **Qwen2.5-VL.** The vision tower in C++; images answered through the server
   and carried across turns. The 50 MB nobody could account for in the weight
   file is an identity matrix. The 35-second vision stage was our CPU
   arithmetic, not the NPU. `src/open_qwen36/vision/`.

6. **PNG decoding.** On a source build, vision models silently ignore PNGs: the
   ffmpeg that vcpkg builds has no zlib, so no PNG decoder, and a failed decode
   is skipped rather than reported. The reader now decodes PNG itself. This
   fixes the closed engine too. `src/common/image/`. There's a design call in
   here I'd like yours on: an own decoder, versus fixing the dependency and the
   installers' DLL list.

7. **Qwen3-VL, including deepstack.** All three image rounds pass. One trap
   worth knowing: `pull` compares files against the manifest's *byte count*, so
   editing a listed config file by a hundred bytes triggers a 4 GB re-download.
   The vision keys this model's config lacks live in a separate file for that
   reason. `src/open_qwen36/vision/`, `core.cpp`.

8. **Dispatch timeouts, diagnosed.** Occasionally an NPU command times out. The
   driver's own record says it was never executed — lost, not late. Not fixed;
   now diagnosable, and the second wait is five seconds instead of sixty.
   `src/open_qwen36/core.cpp`, `cli.cpp`.

9. **GPT-OSS's weight file, read correctly.** No tokens yet; that's #91. Its
   experts are 4-bit float and were being read as integers; its weight blocks
   are half-width and in a different order. Three more silent wrongs: a corrupt
   padded layout, a vocabulary lookup at the wrong stride, a build script that
   misreported failures. And the expert block's memory budget, which it never
   had — it fits at this model's shape, with 4 KB to spare.
   `open_kernels/model/mxfp4.py`, `recipes/pack.py`, `recipes/qwen36moe.py`,
   `src/open_qwen36/pools.cpp`.

**Two things this doesn't fix and that probably deserve issues:** the `pull`
byte-count trap in 7, and the lost NPU command in 8.

**Tests.** 602 pass on the Python side, upstream's own included. `manifest_test`,
`pools_test` and `block_host_test` pass. Qwen2.5, LFM2, Qwen2.5-VL and Qwen3-VL
pass `oflm-test --llm` through `oflm serve`, and both VLMs answer about images.

Fixes #82. Part of #71 and #91.


src/common/image/png_decode.cpp
```diff
@@ -0,0 +1,406 @@
...
+    // The spec's own depth/colour pairings; a file outside them is malformed.
+    if ((color_type == 2 || color_type == 4 || color_type == 6) && depth < 8) {
+        err = "PNG: colour type " + std::to_string(color_type) + " needs 8- or 16-bit samples";
+        return false;
+    }
+    if (color_type == 3 && depth == 16) { err = "PNG: palette images cannot be 16-bit"; return false; }
+
+    // Guard the allocation before doing it: 1 GB of pixels is already far past
+    // anything a vision prompt should carry.
+    const uint64_t stride64 = (static_cast<uint64_t>(w) * nch * depth + 7) / 8;
```
vegah, line 376: This code is prone to buffer overflow, since there are no checks on max width (only min). 
 w = 2^31, h = 2^30, depth = 16, colour_type = 6, interlace = 0
will wrap around and make raw64 == 1<<30
And in unfitler you then risk allocating less bytes. In that example: 
stride = 2^34
In unfilter : 
2^34*2^30 = 2^64 = 0

If I am correct, checking max dimensions is something we should do.

vegah (commented):
Very nice PR! Have looked at some parts of it, and had claude take a look as well.

My main issue is that we only check width and height==0, but not for a max value. I think it is reasonable to check for max values as well. I think I found a buffer overflow bug because of it, but even without it there is no need for extremely large images anyway. 

w=65535, h=5000, depth=8, colour_type=2 is for example below the memory roof, and will allocate ~1gb of memory.

src/common/image/png_decode.cpp
Cyronius: confirmed -- good catch.

vegah (approved):

Atomic-Germ (approved):

PR #95, opened 2026-09-18
open_qwen36_cli: --pmode pins the NPU power mode before a run

### LINKED ISSUE #94: open_qwen36_cli never sets the NPU power mode, so standalone benchmarks aren't comparable
The app sets the NPU's power mode at startup (`src/src/main.cpp`), but the standalone CLI never has. A CLI run just inherits whatever mode the last thing to touch the NPU left behind.

That mode is worth about 2x on this hardware. Three `flm bench` runs of the same closed-engine binary, on the same prompt, minutes apart: 9.05, 17.78 and 16.48 tokens/second of decode. Nothing about the model or the hardware changed between them — only which power mode was still set from before.

That makes any open-vs-closed comparison taken minutes or days apart meaningless unless both sides state what mode they ran in, and today neither does.

Filing this so the gap is recorded; a fix is up next.

Cyronius:
Closes #94.

Adds `--pmode` to `open_qwen36_cli`, matching what the app has always done and the CLI never did: set the NPU's power mode before anything runs. Default is `performance`; `--pmode none` leaves it alone. An invalid mode name is rejected with a clear error instead of silently doing nothing.

Also adds `utilities/bench-pair/bench_pair.py`, a script that runs the open and closed engines against each other on one box in one sitting. It pins the power mode itself, throws away the first (cold) round of each engine, and interleaves the rest — closed, open, closed, open — so whatever the box does over a few minutes hits both sides equally instead of whichever one happened to run first.

## Measured

Pinning the mode and interleaving turned a 2x swing into noise. Three interleaved closed-engine runs, same box, same prompt:

| | spread over 3 runs |
|---|---|
| TTFT | 1.98 / 1.99 / 1.99 s (0.4%) |
| decode | 56.41 / 56.60 / 57.02 ms/token (1.1%) |

Against the 9.05 / 17.78 / 16.48 tokens/second swing quoted in #94, taken on the same binary and prompt without the mode pinned.

## Reviewing

```
python utilities/bench-pair/bench_pair.py --model qwen3:4b \
  --model-dir "%USERPROFILE%\.flm\models\Qwen3-4B-NPU2" \
  --kernels src/xclbins/Qwen3-4B-NPU2/open_kernels \
  --open-cli path/to/open_qwen36_cli.exe --closed "C:/Program Files/flm/flm.exe"
```


Atomic-Germ:
Why did I think it already did that.

Atomic-Germ (approved):

Cyronius:
> Why did I think it already did that.

i don't know, but this has been messing me up for a LONG time with my benchmarks. I couldn't trust any of the numbers!

PR #101, opened 2026-09-19
open_kernels: batched prefill for Gemma 3, and the dx_attn hang it uncovered

### LINKED ISSUE #99: Gemma 3 is the one dense model still prefilling a token at a time
#97 gave every dense model a batched prefill route except Gemma 3. The reason was structural: the engine drove one attention kernel against one position table for the whole model, and Gemma 3 has two layer types — five sliding-window layers for every global one — each needing its own window and its own RoPE table. The sequential path already handles that (`dx` / `dx_local`, `ptab` / `ptab_local`); the block route had no equivalent, so Gemma 3 kept the old one-token-at-a-time prefill.

Gemma 3 also differs from the other six in two things the route's host chain didn't implement: its residual "sandwich" norms (the attention and FFN outputs are normed before they join the residual, with a separate pre-FFN norm) and its GeGLU-tanh activation instead of SiLU.

Measured on this box, Gemma3-4B, all 34 layers, a 1290-token prompt, sequential prefill: **86.6 s**.

Filing this so the gap is recorded; the fix is up next.

### LINKED ISSUE #100: The batched route's attention dispatch hangs on Gemma 3 and Phi-4-mini
Found while putting Gemma 3 on the batched prefill route: the very first `dxB` dispatch times out (ERT state 8), every time, on the first token of the first layer. It is not Gemma 3's sliding-window plumbing — it hangs the same way with the plain global-attention kernel — and it is not transient.

The cause is in `open_kernels/designs/dense/dx_attn.py`, the attention half of the route. It was copied from `dx.py` before `dx.py`'s output handling changed, and it still emits `NHL // HPO` output elements per attention core, where `NHL` is the heads a core owns and `HPO` the heads an output element carries. `dx.py` now emits `min(NHL, HPO)` heads per element and gives every core its own output fifo. When a core owns fewer heads than an element carries, the old formula is **zero**: core 0 never releases an output element, the DMA drain waits for it forever, and the dispatch never completes.

That is exactly Gemma 3's geometry (2 heads per core, 4 per element) and Phi-4-mini's (4 and 8). The six families the route had run on so far all own at least as many heads per core as an element carries, which is why it never showed. Phi-4-mini's route has never actually run on hardware before this — it emits the route (#97) but would hang the same way the moment it was used.

Fix is to bring `dx_attn.py` back in line with `dx.py`; PR to follow. Every dense family's `dx_attn` build changes, so each kernel set needs re-exporting.

Cyronius:
Closes #99. Closes #100.

Gemma 3 was the one dense family #97 left out. The block route drove one attention kernel (`dxB`) against one position table (`ptab`) for the whole model, and Gemma 3's sliding-window layers need their own — the same `dx` / `dx_local` split the sequential path already has. Three changes, in three commits:

**1. The route names its attention kernel per layer type.** `gemm_block` gains `attn_kernel` / `attn_args` (defaulting to today's `dxB` / `ptab`, so every existing manifest parses unchanged) plus `sandwich` / `act`. Gemma 3's `dense_local` layers get a `dxB_local` entry — the same `dx_attn` instruction stream, patched with the 1024-row window — bound to `ptab_local`, and the host chain learns Gemma's sandwich residual norms and GeGLU-tanh alongside the plain SiLU chain. `dense.py`'s `gemm_route()` loops over every layer type instead of refusing anything but one, and fails closed for any family whose activation/norm pairing the host chain doesn't cover.

**2. A real bug in `dx_attn.py`, found the moment Gemma 3 ran.** The first `dxB` dispatch timed out, reproducibly, sliding window or not. `dx_attn.py` had been copied from `dx.py` before `dx.py`'s output handling changed: it emitted `NHL // HPO` output elements per attention core, which is **zero** when a core owns fewer heads than an output element carries — core 0 never releases one, the drain waits forever. Gemma 3 (2 heads per core, 4 per element) and Phi-4-mini (4 and 8) both hit it; the six families the route had run on all have `NHL >= HPO`. `dx_attn.py` now matches `dx.py` (`min(NHL, HPO)` heads per element, one output fifo per core). Every dense family's `dx_attn` build key moves, so each kernel set needs re-exporting.

**3. Spec bookkeeping** — `OPEN-PREFILL-BATCH` and `OPEN-MANIFEST` gain the schema clause and two result paragraphs, and `plans/prefill-batch.md` is archived: with Gemma 3 in, every dense-shaped family in the catalogue emits the route, which was that plan's last step.

## Measured

Same box, same sitting, a 1290-token synthetic prompt (long enough that the sliding window is past its 1024 rows), block route vs. the sequential path. The decode check is the one that matters for a sliding-window family: the block route writes the KV rows `dx_local` decodes against afterward, so a mismatch there would corrupt generation while leaving prefill logits looking fine.

| | greedy tokens after the prefill | prefill logits, all 1290 positions: argmax / top-5 / min corr | prefill, block vs sequential |
|---|---|---|---|
| **Gemma3-4B, all 34 layers** | **8/8 identical** | (6 layers) 1274/1290 · 1230/1290 · 0.99927 | **66.8 s vs 86.6 s (1.30x)** |
| Qwen3-4B, 6 layers — the hd-128 control, on its re-exported set | 32/32 identical | 1276/1290 · 1205/1290 · 0.99989 | 10.8 s vs 19.9 s |
| Phi4-mini, 4 layers — its route's first run ever | 16/16 identical | — | 6.0 s vs 12.1 s |

Every argmax flip on both families is a top-2 margin of a few hundredths of a logit on a random-id prompt — the route's bf16 GEMM, the same behaviour documented for the 35B — and Gemma's profile matches the Qwen3-4B control, so no regression on the hd-128 path. Gemma's 1.30x against Qwen3-4B's 1.66x is expected: past 1024 rows the sequential path's local layers already attend over a capped window, so the route has less to win.

Unit side: `manifest_test.cpp` (defaults on the Qwen3 fixture, the two-kernel route on the Gemma fixture), a new mismatched-pairing guard test in `test_gemma3.py`, all 603 spec tests, and `open_qwen36_cli` compiled against the real XRT SDK.

## Reviewing

Built and run natively on Windows (no WSL) — see the companion docs PR for the toolchain setup, which this repo's docs had wrongly called Linux-only.

```
python open_kernels\export_qwen36_kernels.py --model-dir <Gemma3-4B-NPU2>
open_qwen36_cli --model <dir> --kernels src\xclbins\Gemma3-4B-NPU2\open_kernels --ids-file <1290 ids> --layers 34 --gemm-block --max-tokens 8
open_qwen36_cli --model <dir> --kernels src\xclbins\Gemma3-4B-NPU2\open_kernels --ids-file <1290 ids> --layers 34 --max-tokens 8
```


bong-water-water-bong:
Independent confirmation of the `dx_attn` fix on a shape that hits it: **Qwen3-0.6B** (`NHL = 4`, `HPO = 8`), Strix Halo, Linux + XRT.

- **Before:** with the untouched engine and the kernel set as exported from `main`, `--gemm-block` times out on the first `dxB` dispatch (the driver logs a timed-out job), reproducibly, also from a standalone `pyxrt` dispatch.
- **On this branch:** it runs, and agrees with the sequential path. 300 and 2000 tokens plus 16 greedy decode steps: argmax equal on 17/17 logit vectors at both lengths, min corr 0.9998, max |Δlogit| 0.17. A 2000-token prefill takes 43.4 s on the block route vs 53.8 s sequential.

FYI #105 (draft) builds on this branch and touches `dx_attn.py`.

vegah (approved):

PR #102, opened 2026-09-19
docs: NPU kernel builds work natively on Windows

Cyronius:
Four places in the tree said building the AIE kernels needs Linux or WSL: `utilities/export-kernels.py`'s docstring, `docs/BUILD.md` (which contradicted itself 22 lines later), `src/WinSetup.md`, and the `open-qwen36-kernels` skill. It doesn't. Nothing in `build_design.py` or `export_qwen36_kernels.py` is OS-specific, mlir-aie and Peano both ship `win_amd64` wheels, and mlir-aie's own [`docs/buildHostWinNative.md`](https://github.com/Xilinx/mlir-aie/blob/main/docs/buildHostWinNative.md) gives the native path: download the XRT SDK zip to `C:\Xilinx\XRT`, clone mlir-aie, run `utils\iron_setup.py`, dot-source `iron_env.ps1`. No source build, no WSL. The `open-granite-kernels` skill had already built and run kernel sets that way on 2026-09-06; the other docs were never updated, and they cost me most of an afternoon before I found it.

What is genuinely Linux-only is `utilities/export-kernels.py`'s *own* orchestration — `/opt/xilinx/xrt`, `bin/python`, `lib/python*/site-packages` — so its docstring now says that instead of blaming the build it drives.

Two things worth knowing that the upstream guide doesn't spell out, both hit setting this up today:

- The published Windows `llvm-aie` wheel lacks `llvm-objcopy.exe`, which `iron_setup.py` needs to repair the wheel; it falls back to one on `PATH`. A standalone LLVM install (`winget install LLVM.LLVM`, then its `bin` on `PATH`) satisfies it.
- This tree's designs are written against mlir-aie **1.4.2**'s IRON API (`ironvenv-requirements.txt` says so). Check out the `v1.4.2` tag before running `iron_setup.py`, or it installs the rolling wheel and `dx.py` fails on `import TaskGroup`.

Verified by building the Gemma3-4B, Qwen3-4B and Phi4-mini kernel sets natively and running them (#101).


Cyronius:
Good catch, and all three findings are the same one: I put the v1.4.2 trap in the PR description but not in the docs, which is where someone would actually hit it. Fixed in 46059b9.

Each of the three notes now says to check the mlir-aie clone out at `v1.4.2` before running `iron_setup.py`, and says why it matters: `iron_setup.py` picks its wheel from the checkout, so a clone left on `main` installs the rolling one and `dx.py` dies on `cannot import name 'TaskGroup'`. That is not hypothetical — it is exactly what happened here before the pin was found.

`docs/BUILD.md` also gains the setup its kernel section was assuming you already had (XRT SDK zip to `C:\Xilinx\XRT`, clone, `iron_setup.py`), plus the other undocumented trap: the published Windows `llvm-aie` wheel ships no `llvm-objcopy.exe`, which `iron_setup.py` needs to repair that same wheel, so it needs one on `PATH` (`winget install LLVM.LLVM`).

Atomic-Germ (approved):
I can't directly test this, but if it's passing the build test CI and you're able to build also, it works.

Issue #72, opened 2026-09-12
Whisper needs a kernel we have never written


Cyronius:
Whisper turns speech into text. It is still on the closed engine, and it will not come along for free with the other families, because it does not work like them.

Every model we have opened so far writes one word at a time, and each new word looks back over the words already written. Whisper reads the whole audio clip first, then writes the text while looking back at the audio instead. That looking-back-at-something-else step is the problem: we have no kernel that does it, and it is the core of how Whisper works.

It also does not fit the interface every other engine uses, since it has two halves rather than one. That needs sorting out on the C++ side before any kernel work starts.

New work: the part that reads the audio, the part that looks back at it while writing, the part that writes, and the final layer that picks each word.

Already ours and staying put: everything that happens to the audio file before the model sees it. That is normal C++ on the CPU and does not touch the NPU.

Filing this so it is written down. It is the furthest from anything we have, even though the model itself is small, so it should not be next.

Cyronius:
I would vote this one as lower priority. I think STT and TTS are great uses for the NPU, but it might make sense to have it as a separate push for the next release rather than the one we're prepping for.

on the other hand if someone wants to jump in and tackle it, more power to them

vegah:
I would like to test implementing this, since I need something like this.

PR #103, opened 2026-09-20
feat(open-whisper): the groundwork for Whisper on the open engine

vegah:
First part of Whisper implementation (#72). It does not close the issue; more
  PRs will follow once this is merged.

  The encoder needed no new kernel: every one of its matrix products tiles in the
  bf16 whole-array GEMM the embedding models already use, so seven instruction
  streams over one xclbin cover it.

  This PR:
   - Puts the Whisper engine behind an interface, so open and closed can coexist.
     The closed engine's transcripts are byte-identical to before (18/18).
   - Adds `q4nx-build --open-whisper`, which builds the open engine's weight
     container from the HuggingFace checkpoint.
   - Adds the whisper_gemm kernel set: 7 streams, one xclbin, each within
     rel_fro 7.1e-7 of a float64 reference on hardware.
   - Adds the open encoder (NPU) and decoder (host), gated against transformers
     in float64: the encoder reaches the bf16 ceiling, and 12 of 12 clip/protocol
     token paths match.
   - Lets `serve --asr 1` run on the open engine, selected by
     OFLM_WHISPER_ENGINE or by which files the model directory has. Fixes /v1/audio/transcriptions answering 500 to a request with no file, and adds the oflm-test suite that found it.


vegah:
@Atomic-Germ @Cyronius  - can someone tak a look?

Cyronius (commented):
very nice

Cyronius (approved):

PR #105, opened 2026-09-21
open_kernels: per-core KV planes (KVSLICE=1) for dense attention (draft, stacked on #101)

bong-water-water-bong:
> **Draft, stacked on #101.** The block-prefill route (`dx_attn`) hangs on its first dispatch at `NHL < HPO` (Qwen3-0.6B's shape) without #101's fix, so this was verified on top of it; the diff below therefore also shows #101's commits until it merges. **The change under review is the last commit only** (`git show c9666c3`). It conflicts with #101 in `dx_attn.py` alone, already resolved on this branch.

## Per-core KV planes (`KVSLICE=1`) for dense attention

**Problem.** On Qwen3-0.6B a dense layer costs 0.42 ms at position 0 and grows by about 0.55 ms per 1000 cached rows (1.0 ms at 1024, 2.7 at 4000). About half of that slope is streaming, and it is a dataflow limit, not an arithmetic one: the KV rows reach the four attention cores through **one** `ObjectFifo` produced on one shim tile and broadcast, so the extra cores add compute but no DMA bandwidth. (`ATTN_NULL=1`, which streams but skips the row compute, costs ~0.30 of the ~0.54 µs/row; doubling the fifo depth changes nothing.)

**Change.** Each attention core receives only its own kv heads' rows on its own shim channel.

- The KV buffer becomes `ACORES` planes (plane `c` at `c * KV_BYTES / ACORES`, each row `[K_c | V_c]`), so every window fill stays a single 1-D BD, patched exactly like the row-major one, once per plane. The new row is scattered by one strided drain.
- `dx.py` and `dx_attn.py` take `KVSLICE=1`; `recipes/dense.py` writes `layout.kv_planes` / `kv_plane_bytes`; `KVSLICE` is in `PROBE_VARS`, so it is in the build key (a sliced export is never taken for a plain one).
- Engine: `stream_patch::AttnGeometry::planes` / `AttnPatch::base`; `Core::kv_read_rows` / `kv_write_rows` keep a layout-independent row-major view for checkpoint / restore / `kv_row`; the host full-attention block refuses a planar cache; `--max-ctx` must match the compiled capacity (clear error otherwise).

**A trap worth reviewing.** `attn_row_impl` / `attn_rowb_impl` are `noinline inline`, so a per-TU `-DATTN_KVSLICE` gives two same-named bodies with different indexing and the linker keeps one. The first version of this change built and ran, and was silently wrong for cores 1–3 (the new-row step took the slice's local kv-head index). Renaming the sliced copies overflows the 16 KB core program memory. The routines now take a run-time `bool sl` (window kernels `true`, new-row step `false`): one body everywhere. Explained in the `attn.h` comment and the README.

## Evidence (Qwen3-0.6B, Strix Halo, Linux + XRT, greedy)

- **Logits byte-identical to the plain build** on 9-, 300- and 2000-token prompts, through the sequential prefill **and the block-prefill route** (`--gemm-block`, 300 tokens and 2000 tokens = eight blocks), and across `--twice` (checkpoint/restore), for both exported kernel sets. The decode after a block prefill reads KV rows that `dxB` wrote into the planes, so it exercises both sides.
- Per layer, min ms, control → sliced: 0.97 → 0.77 @1024, 1.53 → 1.14 @2048, 2.65 → 1.79 @4000. Output and activation buffers bit-identical at positions 8 / 101 / 1000; the new row lands correctly in all planes; matches an independent numpy attention reference to 0.2% on all 16 heads.
- Decode 46.9 → 52.7 tok/s at 300 tokens and 20.6 → 27.3–28.0 at 2000. Prefill of 2000 tokens: sequential 53.8 → 42.2 s; block route 43.4 → 31.4 s.
- **Knob unset = no change:** every `insts.bin` (dx, dx_attn, ln, lm_head_q4) and every attention kernel object is byte-identical to the previous build; the manifest layout is identical.
- `manifest_test` gains manifest and patch-table tests (planes 1 and 4, sliding window, wrong number of window fills), and they fail when the plane base is deliberately dropped.

## Not covered

- Only the Qwen3-0.6B geometry (`ACORES = 4`, `RB = 4`, no qkv bias); other models are untried. Shim columns 4.. must have a free MM2S channel.
- Gemma 3's sliding-window layers (`dx_local`) go through the same patch table (`window` is honoured, unit-tested) but were not run with planes.


Cyronius:
adding myself because this relates to some work i've been doing today -- i'll have to see if the efforts will need to be merged.

Cyronius:
1. A sliced set can't be loaded by the server. The engine refuses unless
`--max-ctx` equals the capacity the planes were compiled for, but through
`oflm serve` that number isn't a flag — it comes from the model's
`default_context_length` in the registry (`automodel.cpp:49` into
`engine.cpp:79`), and the manifest's own `max_ctx_default` is never read. A set
exported at the default 4096 therefore throws for any model whose registry entry
says something else, and the error tells the user to set a flag they don't have.
The capacity is fixed by the compiled descriptors anyway, so the engine could
adopt it and warn — which is what `update_max_length` already does when the
requested length is too long.

2. The column budget isn't enforced. Each plane needs a free MM2S channel on its
own shim column, and only columns 4 and up have one, so this is really
`ACORES <= 4`. The assert doesn't say that, `MAX_ATTN_CORES` is 6, and a
six-core family would ask for `Tile(9, 0)`. Granite (5 cores) and Phi4-mini (6)
miss only because their kv-head count doesn't divide — an accident, not a guard.
One more clause in the assert would cover it, and the message should mention the
kv-head divisibility it already checks.

3. The spec hasn't moved. `layout.kv_planes` is a new manifest key with its own
refusal conditions, and `manifest_test.cpp` traces `OPEN-MANIFEST` — that is
where #101 put `attn_kernel`. The knob itself belongs beside `ATTN_FAST` in
`OPEN-ATTN-CONTEXT`'s probe-variable clause.

Atomic-Germ:
@Cyronius  #101 is pulled

